PR #108, opened 2025-06-23
feat(cli): add option for infer bindings for init

gheorghitahurmuz:


src/uipath_langchain/_cli/cli_init.py
```diff
@@ -93,8 +93,27 @@ def generate_schema_from_graph(graph: CompiledStateGraph) -> Dict[str, Any]:
...
+
+    try:
+        with open(config_path, "r") as config_file:
+            existing_config = json.load(config_file)
+            return existing_config.get("settings")
+    except (json.JSONDecodeError, IOError):
+        return None
+
+
+async def langgraph_init_middleware_async(entrypoint: str, infer_bindings: bool = True) -> MiddlewareResult:
```
radu-mocanu, line 116: we are not passing the infer_bindings parameter where this function si invoked https://github.com/UiPath/uipath-langchain-python/pull/108/files#diff-8669e540b78a92e73869eead71d817265ad7533011a5675271aa552adf62d076R224
However that would be a breaking change

PR #188, opened 2025-09-22
[WIP] Preconfigured Tools

akshaylive:


pyproject.toml
```diff
@@ -17,6 +17,8 @@ dependencies = [
     "httpx>=0.27.0",
     "openai>=1.65.5",
     "openinference-instrumentation-langchain>=0.1.50",
+    "jsonschema-pydantic>=0.6",
```
cristipufu, line 20: do we actually need this? we could just pass the schema as `dict[str, Any]` directly from the `agent.json`:
```python
ArgsSchema = Union[TypeBaseModel, dict[str, Any]]
```

src/uipath_langchain/tools/preconfigured.py
```diff
@@ -0,0 +1,178 @@
...
+
+def create_escalation_tool_from_channel(channel: AgentEscalationChannel) -> BaseTool:
+    async def escalate(**kwargs) -> BaseModel:
+        return interrupt(
+            CreateAction(
+                title=channel.description,
+                data=kwargs,
+                assignee=None,  # It does not make sense to expose the recipient choice to the agent even though it's technically possible.
+                app_name=channel.properties.app_name,
+                app_folder_path=channel.properties.folder_name,
```
cristipufu, line 54: folder name is not the folder path (the folder path represents the friendly name used to identify a folder and it contains the whole folder hierarchy)

src/uipath_langchain/tools/preconfigured.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+
+
+def create_escalation_tool_from_channel(channel: AgentEscalationChannel) -> BaseTool:
+    async def escalate(**kwargs) -> BaseModel:
+        return interrupt(
+            CreateAction(
+                title=channel.description,
+                data=kwargs,
+                assignee=None,  # It does not make sense to expose the recipient choice to the agent even though it's technically possible.
```
cristipufu, line 52: the asignee can be specified at design-time (so we should try to get it from the agent.json)

src/uipath_langchain/tools/preconfigured.py
```diff
@@ -0,0 +1,178 @@
...
+            return create_cached_wrapper(create_escalation_tool(resource), cache)
+        case _:
+            raise NotImplementedError()
+
+
+logger = logging.getLogger(__name__)
+
+
+def safe_extract_tools(
+    agent_definition: AgentDefinition, cache: Optional[SQLiteCache] = None
```
cristipufu, line 169: should we use BaseCache (instead of SQLiteCache)?

pyproject.toml
```diff
@@ -17,6 +17,8 @@ dependencies = [
...
+    "types-requests>=2.32.4.20250913",
```
cristipufu, line 21: use `httpx`

pyproject.toml
akshaylive: This one is for

```
OutputType: Type[BaseModel] = create_model(resource.output_schema)
```

Having a stronger `OutputType` definition will help LLMs invoke tools as intended.

src/uipath_langchain/tools/preconfigured.py
akshaylive: Channels do not specify folder paths. I've changed this to `None`. Please resolve this conversation if that assumption is sufficient.

PR #224, opened 2025-10-16
feat: update to latest uipath-python model changes to lowcode agent

cotovanu-cristian:


cotovanu-cristian:
reverted these changes in `uipath-python` to avoid package breaking changes without major version bump.
closing PR till we do the renaming and version bump

PR #231, opened 2025-10-21
feat: refactor cli internals, add langgraph runtime factory

bai-uipath:
This PR adds a few runtime methods needed for dependencies:
* Added a `LangGraphRuntimeFactory` for easier runtime construction outside this package
* Added a `get_entrypoints()` method to enable generating entrypoint schemas
* Minor refactor to `cli_eval.py` to use the new evaluate() helper

### DIFF
--- a/src/uipath_langchain/_cli/_runtime/_runtime.py
+++ b/src/uipath_langchain/_cli/_runtime/_runtime.py
@@ -2,12 +2,14 @@ import logging
 import os
 from contextlib import asynccontextmanager
 from typing import Any, AsyncGenerator, AsyncIterator, Optional, Sequence
+from uuid import uuid4
 
 from langchain_core.runnables.config import RunnableConfig
 from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver
 from langgraph.errors import EmptyInputError, GraphRecursionError, InvalidUpdateError
 from langgraph.graph.state import CompiledStateGraph, StateGraph
 from langgraph.types import Interrupt, StateSnapshot
+from typing_extensions import override
 from uipath._cli._runtime._contracts import (
     UiPathBaseRuntime,
     UiPathBreakpointResult,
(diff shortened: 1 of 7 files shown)


src/uipath_langchain/_cli/_utils/_schema.py
```diff
@@ -0,0 +1,87 @@
+"""Schema extraction utilities for LangGraph graphs."""
+
+from typing import Any, Dict
+
+from langgraph.graph.state import CompiledStateGraph
+
+
+def resolve_refs(schema, root=None):
```
bai-uipath: Note: these functions are moved from `cli_init.py` without modification

src/uipath_langchain/_cli/cli_init.py
```diff
@@ -1,230 +0,0 @@
...
-        if write_config:
-            config_path = write_config(uipath_config)
-        else:
-            # Save the uipath.json file
-            config_path = "uipath.json"
-            with open(config_path, "w") as f:
-                json.dump(uipath_config, f, indent=4)
-
-        for graph_name, mermaid_content in mermaids.items():
-            mermaid_file_path = f"{graph_name}.mermaid"
```
bai-uipath: note: I've removed the mermaid file generation since we're no longer doing custom logic in the `uipath init`. These files don't appear to be used anywhere.

src/uipath_langchain/_cli/_runtime/_runtime.py
```diff
@@ -475,6 +482,25 @@ def __init__(
         self.resolver = LangGraphJsonResolver(entrypoint=entrypoint)
         super().__init__(context, self.resolver)
 
+    @cached_property
+    @override
+    def get_entrypoint(self) -> Entrypoint:
+        """Get entrypoint for this LangGraph runtime."""
+        with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
```
cristipufu: I think we can just make this async in the base library
```python
async def get_entrypoint(self) -> Entrypoint:
```

Also it looks like the resolved graph is already cached in `LangGraphJsonResolver` (so `@cached_property` doesn't bring much value here)

PR #304, opened 2025-12-03
feat(StructuredToolWithOutputType): support dynamic output types

akshaylive:
`StructuredTool`s are `Runnable`s, but `OutputType` is always `Any` unless overridden. This commit adds the override since we have strongly typed output types.

Additionally, removing asterisks from IS output schemas (bugfix).

### DIFF
--- a/src/uipath_langchain/agent/tools/integration_tool.py
+++ b/src/uipath_langchain/agent/tools/integration_tool.py
@@ -12,6 +12,7 @@ from uipath.platform import UiPath
 from uipath.platform.connections import ActivityMetadata, ActivityParameterLocationInfo
 
 from .static_args import handle_static_args
+from .structured_tool_with_output_type import StructuredToolWithOutputType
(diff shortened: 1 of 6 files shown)


src/uipath_langchain/agent/tools/integration_tool.py
```diff
@@ -143,11 +185,14 @@ async def integration_tool_fn(**kwargs: Any):
...
-    tool = StructuredTool(
+    tool = StructuredToolWithOutputType(
         name=tool_name,
         description=resource.description,
         args_schema=input_model,
         coroutine=integration_tool_fn,
+        output_type=output_model,
     )
 
+    tool.__dict__["OutputType"] = output_model
```
radu-mocanu: is this needed if the output_type property is already overriden?

radu-mocanu:
other than the comment above LGTM

radu-mocanu (approved):

PR #322, opened 2025-12-08
feat: add support for HITL tools [AL-227]

ctiliescuuipath:


src/uipath_langchain/agent/react/agent.py
```diff
@@ -72,7 +72,7 @@ def create_agent(
     llm_tools: list[BaseTool] = [*agent_tools, *flow_control_tools]
 
     init_node = create_init_node(messages)
-    tool_nodes = create_tool_node(agent_tools)
+    tool_nodes = create_tool_node(agent_tools, guardrails)
```
valentinabojan: Can we maybe decorate the tool_nodes with guardrails? Instead og adding guardrails concept into the create_tool_node function?
Something like:
```
tool_nodes = create_tool_node(agent_tools)
tool_nodes_with_guardrails = create_tool_guardrails_subgraph(tool_nodes, ...)
```

src/uipath_langchain/agent/guardrails/actions/escalate_action.py
```diff
@@ -200,6 +230,82 @@ def _process_escalation_response(
...
+                return {}
+
+            # Update tool calls with reviewed args
+            if last_message.tool_calls and reviewed_tool_calls_args:
+                tool_calls = list(last_message.tool_calls)
+                for i, tool_call in enumerate(tool_calls):
+                    if i < len(reviewed_tool_calls_args):
+                        reviewed_args = reviewed_tool_calls_args[i]
+                        if isinstance(reviewed_args, dict):
+                            if isinstance(tool_call, dict):
```
valentinabojan: How this works? When can we have multiple tool calls on the message for tool nodes?

PR #331, opened 2025-12-10
feat: manual instrumentation for span hierarchy parity with low-code agents

saksharthakkar:
## Problem Statement

The current OpenInference auto-instrumentation for LangGraph agents has two significant issues:

1. **Large trace payload size** - Auto-instrumentation produces excessive spans including internal LangGraph operations (init, route_agent, terminate, etc.). For a simple agent with one tool call, auto-instrumentation generates **~50+ spans** vs **~7 spans** with manual instrumentation, resulting in trace files **5-10x larger** than necessary.

2. **Not user-friendly** - The generated span structure doesn't match UiPath's expected schema, making traces hard to correlate with Temporal implementation (low-code agents) and difficult to visualize in UiPath trace viewers.

## Solution

Implement manual instrumentation to achieve trace schema parity with the Temporal implementation. This creates clean, predictable span hierarchies that match exactly what low-code agents produce.

## Scope of This PR

This PR implements **span hierarchy parity** for:
- Agent run (`agentRun`)
- LLM call (`completion`) / Model run (`llmCall`) 
- Tool call (`toolCall`)
- Agent output (`agentOutput`)

**Note:** Full span attributes and other nodes support(e.g. guardrails) will be implemented in follow-up PRs.

## UIPATH_CUSTOM_INSTRUMENTATION Env Variable

All changes are gated behind `UIPATH_CUSTOM_INSTRUMENTATION` environment variable:
- **Default: `false`** (disabled) - uses existing OpenInference auto-instrumentation
- Set to `"true"` to enable manual instrumentation

### New Files
- `src/uipath_langchain/_tracing/schema.py` - Span types and names matching Temporal implementation schema
- `src/uipath_langchain/_tracing/tracer.py` - `UiPathTracer` class with context managers for span creation
- `tests/tracing/test_tracer.py` - Unit tests for tracer (12 tests)

### Modified Files
- `src/uipath_langchain/runtime/runtime.py` - Wraps `execute()` and `stream()` in Agent run span
- `src/uipath_langchain/agent/react/llm_node.py` - Adds LLM call + Model run nested spans
- `src/uipath_langchain/agent/tools/tool_node.py` - Adds Tool call spans

## Test Summary

**Unit Tests:** 12 tests covering:
- Agent run span creation and attributes
- LLM call span with correct type (`completion`)
- Model run span with model name attribute
- Tool call span with tool name
- Agent output span
- Full span hierarchy (parent-child relationships)
- Error handling and status propagation
- Tracer singleton behavior

## Manual Test Summary

Tested locally using CLI in `uipath-agents-python` repo with basic example:
```bash
export UIPATH_CUSTOM_INSTRUMENTATION=true
uv run uipath run agent.json --trace-file traces.jsonl
```
**With openinference instrumentation**
image

**With custom instrumentation**
image

Note: 2 LLM calls is based on current [implementation](https://uipath-product.slack.com/archives/C09Q8B4FYM8/p1765350932975909?thread_ts=1765301252.657909&cid=C09Q8B4FYM8) 

```

## Next Steps (Future PRs)
Attributes and other nodes parity, upsertSpans to update running/completed states.
There are 40+ attributes classes in current implementation in Agents repo.
<img width="700" height="945" alt="image" src="https://github.com/user-attachments/assets/a4fcb38a-6126-4a68-ab02-649d2036448d" />

### DIFF
--- a/src/uipath_langchain/_tracing/schema.py
+++ b/src/uipath_langchain/_tracing/schema.py
@@ -3,6 +3,7 @@
 This module defines the span types and naming conventions that match
 the Temporal implementation backend.
 """
+
 from enum import Enum
 
 
--- a/src/uipath_langchain/_tracing/tracer.py
+++ b/src/uipath_langchain/_tracing/tracer.py
@@ -3,6 +3,7 @@
 This module provides the UiPathTracer class that creates OpenTelemetry spans
 with the exact schema matching Temporal implementation backend.
 """
+
 import json
 import os
 from contextlib import contextmanager
--- a/tests/tracing/test_tracer.py
+++ b/tests/tracing/test_tracer.py
@@ -1,4 +1,5 @@
 """Test UiPathTracer produces correct span structure."""
+
 import json
 
 import pytest

### CODE REVIEW

**Review by cristipufu — CHANGES_REQUESTED** [review]
Move everything to the low-code repo, use the factory's `setup_instrumentation` method to override the standard LangGraph instrumention

File: src/uipath_langchain/agent/react/llm_node.py [hunk 1]
```diff
@@ -32,12 +41,41 @@ async def llm_node(state: AgentGraphState):
         else:
             llm = base_llm
 
+        if is_custom_instrumentation_enabled():
+            return await _llm_node_instrumented(llm, messages, model)
+
+        # Original behavior - OpenInference handles tracing
```
smflorentino: we should not duplicate implementations just to instrument spans

src/uipath_langchain/agent/react/llm_node.py
```diff
@@ -32,12 +41,41 @@ async def llm_node(state: AgentGraphState):
...
+    Creates nested span hierarchy matching C# Agents:
+    - LLM call (outer, type: completion) - wraps the iteration
+    - Model run (inner, type: llmCall) - actual API call with model name
+    """
+    tracer = get_tracer()
+
+    # Extract model name from the model instance
+    model_name = getattr(model, "model_name", getattr(model, "model", "unknown"))
+
+    with tracer.start_llm_call():
```
smflorentino: this doesn't handle setting attributes after the LLM call completes

src/uipath_langchain/runtime/runtime.py
```diff
@@ -61,7 +65,44 @@ async def execute(
...
+        input: dict[str, Any] | None = None,
+        options: UiPathExecuteOptions | None = None,
+    ) -> UiPathRuntimeResult:
+        """Execute with custom instrumentation - wraps in Agent run span."""
+        tracer = get_tracer()
+
+        # Use entrypoint as agent name, fallback to runtime_id
+        agent_name = self.entrypoint or self.runtime_id
+
+        with tracer.start_agent_run(
```
smflorentino: why is the start called `start_agent_run` and the finish called `emit_agent_output`...shouldn't the methods be named more similarly, to represent opening and closing of spans? 
such as `start_agent_run` and `finish_agent_run`

that said manually writing start/finish methods for every type of span doesn't seem ideal. I think we need a way to wrap an async method, emit a span *before* the method, close it after it finishes (including handling errors).

we probably need to sprinkle logic (manually) cross the low-code codebase but:
- we should see how to wrap arbitrary nodes rather than editing each node (address @cristipufu concern)
- have common helper methods to wrap start/stop so callers only need to specify attribute extraction logic
- figure out how to handle the `OpenInference` [issue](https://uipath-product.slack.com/archives/C09GGFG1Q56/p1765369088945959) (this can presumably come after the first two)

saksharthakkar:
Closing this PR as per comments and offline discussions, will move these changes to agents-python repo... might make changes in future PR to add hooks to support nested hierarchy for traces if needed.

PR #343, opened 2025-12-12
feat(prompts_sample): add sample for mcp prompts

giuliastf:
The main purpose of this pr s to show how can be used the available ```get_prompt()``` LanGraph function for MCP servers.


samples/mcp-refactoring-assistant/graph.py
```diff
@@ -0,0 +1,179 @@
...
+    user_msg = HumanMessage(content=state.code)
+
+    messages_state = MessagesState(messages=[system_msg, user_msg])
+    result = await react_agent.ainvoke(messages_state)
+
+    prompt_name = ""
+    prompt_args: dict = {}
+
+    for msg in result.get("messages", []):
+        if hasattr(msg, 'type') and msg.type == 'tool':
```
radu-mocanu: the logic is a bit hard to follow here (in general samples should be easy to understand). what if we instruct the agent to output for us the `prompt_name` and `arguments` so we don't need to search for them in the list of messages.

```
class StructuredOutput(BaseModel):
   """Description here"""
   prompt_name: str = Field(description="description")
    arguments: str = Field(description="description")
```

ref from docs: https://docs.langchain.com/oss/python/langchain/agents#toolstrategy.

then we just extract the putput from result["structured_output"] and already have access to what we are interested in.

PR #347, opened 2025-12-15
feat: integrate deterministic guardrails [AL-229]

ctiliescuuipath:
1. convert AgentCustomGuardrails to DeterministicGuardrails
2. integrate them in guardrail node

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,40 @@
+# Copilot Instructions for Review
+
+Purpose:
+- Always use `**get_httpx_client_kwargs()` when creating `httpx.Client` or `httpx.AsyncClient` to ensure correct SSL, proxy, timeouts, and redirects.
+
+Guidance for Copilot:
+- When suggesting `httpx.Client(` or `httpx.AsyncClient(`, include `**get_httpx_client_kwargs()`.
+- If the user omits it, suggest the corrected version or insert a warning comment.
+
+Preferred patterns:
+```python
+import httpx
+from uipath._utils._ssl_context import get_httpx_client_kwargs
+
+client = httpx.Client(**get_httpx_client_kwargs())
+async_client = httpx.AsyncClient(**get_httpx_client_kwargs())
+
+default_client_kwargs = get_httpx_client_kwargs()
+client_kwargs = {
+    **default_client_kwargs,
+    "base_url": base_url,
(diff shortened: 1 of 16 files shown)


src/uipath_langchain/agent/guardrails/guardrails_factory.py
```diff
@@ -19,6 +34,177 @@
...
+        operator: The word operator to convert.
+        value: The value to compare against (may be None for isEmpty/isNotEmpty).
+
+    Returns:
+        A callable that takes a string and returns a boolean.
+    """
+    match operator:
+        case AgentWordOperator.CONTAINS:
+            assert value is not None, "value cannot be None for CONTAINS operator"
+            return lambda s: value in s
```
valentinabojan: Is it case insensitive? I think in .NET it was case insensitive

src/uipath_langchain/agent/guardrails/utils.py
```diff
@@ -1,4 +1,146 @@
...
+        return {}
+
+    # For PRE_EXECUTION, look at last message
+    # For POST_EXECUTION, look at second-to-last message (before the ToolMessage)
+    if execution_stage == ExecutionStage.PRE_EXECUTION:
+        if len(state.messages) < 1:
+            return {}
+        message = state.messages[-1]
+    else:  # POST_EXECUTION
+        if len(state.messages) < 2:
```
valentinabojan: @cotovanu-cristian Do we already have some functions to reuse here and get the tool input/output? I would like to avoid re-implementing something that we have (if the case).

src/uipath_langchain/agent/guardrails/utils.py
```diff
@@ -1,4 +1,146 @@
...
+
+    Args:
+        message: The message to extract from.
+        tool_name: Name of the tool to extract arguments from.
+
+    Returns:
+        Dict containing tool call arguments, or empty dict if not found.
+    """
+    if not isinstance(message, AIMessage):
+        return {}
```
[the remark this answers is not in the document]
ctiliescuuipath: no, but to be uniform i will return {} also here https://github.com/UiPath/uipath-langchain-python/blob/2e8e2a96cbda682efeda609d38be00c771d2adb6/src/uipath_langchain/agent/guardrails/actions/escalate_action.py#L448-L448

PR #387, opened 2026-01-05
feat: add asset recipient support for escalations

dianagrecu-uipath:
## Description
Add support for Asset Recipient in Escalations

## Testing
Locally and added unit tests

### DIFF
--- a/src/uipath_langchain/agent/tools/escalation_tool.py
+++ b/src/uipath_langchain/agent/tools/escalation_tool.py
@@ -32,10 +32,10 @@ class EscalationAction(str, Enum):
     END = "end"
 
 
-def resolve_recipient_value(recipient: AgentEscalationRecipient) -> str | None:
+async def resolve_recipient_value(recipient: AgentEscalationRecipient) -> str | None:
     """Resolve recipient value based on recipient type."""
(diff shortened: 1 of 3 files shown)


src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -29,11 +31,41 @@ class EscalationAction(str, Enum):
...
+    if (
+        recipient.type == AgentEscalationRecipientType.ASSET_USER_EMAIL
+        or recipient.type == AgentEscalationRecipientType.ASSET_GROUP_NAME
+    ):
+        return resolve_asset(recipient.asset_name, recipient.folder_path)
+
+    if hasattr(recipient, "value"):
+        return recipient.value
+
+    return None
```
cotovanu-cristian: let's use type narrowing here as `AgentEscalationRecipient` is a discriminated union and can be used specifically for such scenarios.


```suggestion
    if isinstance(recipient, AssetRecipient):
        return resolve_asset(recipient.asset_name, recipient.folder_path)

    if isinstance(recipient, StandardRecipient):
        return recipient.value

    return None
```

cotovanu-cristian (approved):

PR #389, opened 2026-01-07
fix(ToolNode): pass correct inputs to tool nodes

akshaylive:
Scenario:

1. LLM invokes multiple instances of the same tool, such as ["Web Search", "Web Search"]
1. [This line](https://github.com/UiPath/uipath-langchain-python/blob/main/src/uipath_langchain/agent/react/router.py#L89) schedules multiple tool nodes to run in parallel.
1. Each tool node picks up the [first matching tool call](https://github.com/UiPath/uipath-langchain-python/blob/main/src/uipath_langchain/agent/tools/tool_node.py#L86-L88) from the state. This means that both of the web search tools will execute the first tool call.

Fix is to send the correct `ToolCall` details to the tool node. This also simplifies the tool node implementation.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-langchain==0.3.3.dev1003891788",

  # Any version from PR
  "uipath-langchain>=0.3.3.dev1003890000,<0.3.3.dev1003900000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```

### DIFF
--- a/src/uipath_langchain/agent/tools/tool_node.py
+++ b/src/uipath_langchain/agent/tools/tool_node.py
@@ -4,13 +4,14 @@ from collections.abc import Sequence
 from inspect import signature
 from typing import Any, Awaitable, Callable, Literal
 
-from langchain_core.messages.ai import AIMessage
 from langchain_core.messages.tool import ToolCall, ToolMessage
 from langchain_core.runnables.config import RunnableConfig
 from langchain_core.tools import BaseTool
 from langgraph._internal._runnable import RunnableCallable
 from langgraph.types import Command
 from pydantic import BaseModel
+
+from ..react.types import AgentGraphState
 
 # the type safety can be improved with generics
 ToolWrapperType = Callable[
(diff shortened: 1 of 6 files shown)


andreitava-uip (changes requested):
The design is overall what I was going for as well, however this PR has 2 issues:
1. We still need to pass the entire state to the node, even with the send API. The wrapper requires the state to be passed as many of them rely on it (static args wrapper, job attachments wrapper)
2. This is more of a nit, but It would be nice for our ToolNode to support both direct ToolCalls and entire state passing. These changes as they are now force anybody wanting to use the node into the Send API.

src/uipath_langchain/agent/tools/tool_node.py
```diff
@@ -61,7 +64,7 @@ def _func(self, state: Any, config: RunnableConfig | None = None) -> OutputType:
         return self._process_result(call, result)
 
     async def _afunc(
-        self, state: Any, config: RunnableConfig | None = None
+        self, state: AgentGraphState, config: RunnableConfig | None = None
```
andreitava-uip: do not make this AgentGraphState.
We need the entire state here, including the input args, for which there is the runtime dynamically created type `CompleteAgentGraphState`. Unfortunately since it's created at runtime we cannot annotate with it.

PR #395, opened 2026-01-12
feat: conversational agent support

mike-deem-uipath:
Depends on https://github.com/UiPath/uipath-python/pull/1087 and https://github.com/UiPath/uipath-runtime-python/pull/65 (is causing checks to fail)

Adds support for conversation agents:
* state is initialized with input messages received when the job is started. init node overwrites messages in state so that the system prompt message comes first.
* no tools are bound, the agent can just respond without invoking a tool
* only agent response messages are sent to cas (or both coded and autonomous conversational agents). CAS doesn't expect the system message and the user message is sent from the client.
*  don't delete state.db if keep_state_file context option is set (the cli can set this using --keep-state-file). This allows conversational agents to accumulate conversation history across multiple runs.

### DIFF
--- a/src/uipath_langchain/agent/react/router_conversational.py
+++ b/src/uipath_langchain/agent/react/router_conversational.py
@@ -1,0 +1,43 @@
+"""Routing functions for conditional edges in the agent graph."""
+
+import logging
+from typing import Literal
+
+from uipath_langchain.agent.react.router_utils import validate_last_message_is_AI
+
+from .types import AgentGraphNode, AgentGraphState
+
+logger = logging.getLogger(__name__)
+
+
+def create_route_agent_conversational():
+    """Create a routing function for conversational agents. It routes between agent and tool calls until
+    the agent response has no tool calls, then it routes to the USER_MESSAGE_WAIT node which does an interrupt.
+
+    Returns:
+        Routing function for LangGraph conditional edges
+    """
+
+    def route_agent_conversational(
+        state: AgentGraphState,
+    ) -> list[str] | Literal[AgentGraphNode.TERMINATE]:
+        """Route after agent
+
+        Routing logic:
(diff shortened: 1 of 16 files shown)


src/uipath_langchain/agent/react/router_conversational.py
```diff
@@ -0,0 +1,58 @@
...
+        last_message = validate_last_message_is_AI(state.messages)
+        if last_message.tool_calls:
+            return [tc["name"] for tc in last_message.tool_calls]
+        else:
+            return AgentGraphNode.USER_MESSAGE_WAIT
+
+    return route_agent_conversational
+
+
+def create_user_message_wait_node():
```
cristipufu: We should just route to terminate / end the execution of the graph (we don't need a new resume trigger)

mike-deem-uipath: done

src/uipath_langchain/chat/mapper.py
```diff
@@ -307,21 +307,8 @@ def map_event(
...
-                start=UiPathConversationMessageStartEvent(
-                    role="assistant", timestamp=timestamp
-                ),
-                content_part=UiPathConversationContentPartEvent(
-                    content_part_id=f"cp-{message.id}",
-                    chunk=UiPathConversationContentPartChunkEvent(data=text_content),
-                ),
-                end=UiPathConversationMessageEndEvent(),
-            )
-        ]
```
andreitava-uip: is it safe to remove these? I don't exactly understand the impact

andreitava-uip (approved):
Overall it looks ok.
Some things will need to be changed to properly support parallel execution of tools, but that's my job

cotovanu-cristian (approved):

PR #398, opened 2026-01-12
feat: add tool output job attachments in the agent state

cristian-groza:
- Add tool output job attachments in the agent state

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-langchain"
-version = "0.3.2"
+version = "0.3.3"
 description = "Python SDK that enables developers to build and deploy LangGraph agents to the UiPath Cloud Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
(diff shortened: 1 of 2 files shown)


tests/agent/wrappers/test_job_attachment_wrapper.py
```diff
@@ -573,3 +571,156 @@ async def test_tool_with_complex_nested_structure_all_valid(
...
+            str(output_attachment_id) in result.update["inner_state"]["job_attachments"]
+        )
+        assert (
+            result.update["inner_state"]["job_attachments"][str(output_attachment_id)]
+            == output_attachment
+        )
+        # Verify get_job_attachments was called with correct parameters
+        mock_get_job_attachments.assert_called_once_with(
+            MockOutputSchema, mock_tool.ainvoke.return_value
+        )
```
andreitava-uip: why don't we reuse `assert_command_success` here? Logic seems similar enough to me

PR #404, opened 2026-01-13
feat: add compatibility for guardrails entitlements phase 1

apetraru-uipath:
Added compatibility for guardrails entitlements so that tests are passing.

### DIFF
--- a/src/uipath_langchain/agent/guardrails/guardrail_nodes.py
+++ b/src/uipath_langchain/agent/guardrails/guardrail_nodes.py
@@ -100,10 +100,28 @@ def _create_validation_command(
     Returns:
         Command to update state and route to appropriate node.
     """
-    if not result.validation_passed:
+    from uipath.core.guardrails import GuardrailValidationResultType
+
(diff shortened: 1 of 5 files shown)


src/uipath_langchain/agent/guardrails/guardrail_nodes.py
```diff
@@ -100,10 +100,28 @@ def _create_validation_command(
...
+    from uipath.core.guardrails import GuardrailValidationResultType
+
+    # Handle new format: result has a 'result' attribute with GuardrailValidationResultType
+    if hasattr(result, "result"):
+        if result.result != GuardrailValidationResultType.PASSED:
+            return Command(
+                goto=failure_node, update={"guardrail_validation_result": result.reason}
+            )
+    # Handle old format: backwards compatibility for tests using validation_passed
+    elif hasattr(result, "validation_passed"):
```
valentinabojan: Do we need this check if we don't mind backwards compatibility for now?

PR #425, opened 2026-01-18
feat: add file support for more LLM API Flavors

ionut-mihalache-uipath:
Add LLM Provider and API Flavor Support for File Handling

Summary

- Introduces LLMProvider and APIFlavor enums to explicitly identify LLM vendors and their API variations
- Adds llm_provider and api_flavor attributes to all UiPath LLM classes (UiPathChatOpenAI, UiPathChatBedrockConverse, UiPathChatBedrock, UiPathChatVertex)
- Refactors file content part building to use provider-specific message formats based on API flavor


src/uipath_langchain/chat/openai.py
```diff
@@ -64,6 +65,9 @@ def handle_request(self, request: httpx.Request) -> httpx.Response:
 
 
 class UiPathChatOpenAI(AzureChatOpenAI):
+    llm_provider: LLMProvider = LLMProvider.OPENAI
+    api_flavor: APIFlavor = APIFlavor.OPENAI_RESPONSES
```
cotovanu-cristian: if we choose this solution of exposing these metadata fields on the class it should be computed based on the same logic that the underlying`BaseChatOpenAI` class uses.

1. it is either set through the `use_responses_api` during model instantiation

Image

OR 

2. check the values of other params and determines based on their structure

Image

Image

cotovanu-cristian (commented):
the provider detection functionality should be extract in a new common module `/llm/` that should expose these common provider specific operations.
e.g: `get_llm_provider()` | `get_api_flavor()`

src/uipath_langchain/agent/react/llm_node.py
```diff
@@ -21,15 +23,22 @@
...
+        - {"type": "any"} for Bedrock Invoke API (dict format required)
     """
     model_class_name = model.__class__.__name__
     if model_class_name in OPENAI_COMPATIBLE_CHAT_MODELS:
         return "required"
+
+    api_flavor = getattr(model, "api_flavor", None)
+    if api_flavor == APIFlavor.AWS_BEDROCK_INVOKE:
+        return {"type": "any"}
+
```
andreitava-uip: nit: I'd rather explicitly store model info somewhere, instead of relying on props on `BaseChatModel`.
My main argument is that we will require this information in many other places (such as tool nodes, guardrail nodes and so on), many of which will not have access to the `BaseChatModel` instance.

I think we should explore passing them via State/RuntimeContext/RunnableConfig

src/uipath_langchain/chat/__init__.py
```diff
@@ -29,14 +29,21 @@ def __getattr__(name):
         from . import supported_models
 
         return getattr(supported_models, name)
+    if name in ("LLMProvider", "APIFlavor", "UiPathChatModel"):
+        from . import types
+
+        return getattr(types, name)
```
andreitava-uip: why do we need this?

src/uipath_langchain/llm/builders/base.py
```diff
@@ -0,0 +1,56 @@
...
+        """Build a provider-specific content part for a file.
+
+        Args:
+            url: URL to download the file from.
+            filename: Name of the file.
+            mime_type: MIME type of the file.
+
+        Returns:
+            Provider-specific content part dictionary.
+        """
```
andreitava-uip: [Discussion]:
Do we want a higher level of abstraction here?
I was thinking that these Message Adapters/Handler/Builders would work with Messages and return messages, not content parts.
For example have a `set_files_in_content(message, files) -> message` that both creates the parts and inserts/replaces in the content.

My concern is that even if we do get the content parts, actually inserting/replacing them to content properly might still be API specific

src/uipath_langchain/chat/models.py
```diff
@@ -169,6 +174,9 @@ def endpoint(self) -> str:
 class UiPathChat(UiPathRequestMixin, AzureChatOpenAI):
     """Custom LLM connector for LangChain integration with UiPath Normalized."""
 
+    llm_provider: LLMProvider = LLMProvider.OPENAI
+    api_flavor: APIFlavor = APIFlavor.OPENAI_COMPLETIONS
```
ionut-mihalache-uipath: I will delete llm_provider and api_flavor from here, because UiPathChat is the class used for normalized apis. Normalized API will no longer be supported for agents.

cotovanu-cristian (approved):

PR #431, opened 2026-01-19
feat: add static args support

andreiancuta-uipath:


src/uipath_langchain/agent/tools/schema_editing.py
```diff
@@ -0,0 +1,234 @@
...
+) -> dict[str, Any]:
+    """Enters an anyOf union and inlines the ref if present.
+
+    Args:
+        schema: The root schema containing $defs
+        container: The container of the anyOf, modified in place
+
+    Returns:
+        The non-null schema of the anyOf, inlined if needed. If there is no anyOf, returns the container.
+    """
```
andreitava-uip: so this doesn't support real `anyOf/oneOf` between multiple schemas?
Pretty sure there are some IS tools that have them (and others use `if then else`)
Speaking of, what about conditional schemas with `if then else`?

src/uipath_langchain/agent/tools/static_args.py
```diff
@@ -113,4 +252,5 @@ def handle_static_args(
 
     static_args = resolve_static_args(resource, dict(state))
     merged_args = apply_static_args(static_args, input_args)
-    return merged_args
+    sanitized_merged_args = sanitize_dict_for_serialization(merged_args)
```
andreitava-uip: why do we sanitize again? do we insert basemodels?

andreitava-uip (approved):

PR #448, opened 2026-01-21
Feat/add guardrails task recipient

dianapirvulescu:
Add support for `TaskRecipient` to be used in an escalation action.

### DIFF
--- a/docs/quick_start.md
+++ b/docs/quick_start.md
@@ -171,26 +171,6 @@ This command creates the following files:
 | `bindings.json`     | Allows you to configure overridable resource bindings                        |
 | `agent.mermaid`     | Graph visual representation                                                  |
 
-## Set Up Environment Variables
-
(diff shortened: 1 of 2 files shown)


src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -32,13 +34,24 @@ class EscalationAction(str, Enum):
     END = "end"
 
 
-async def resolve_recipient_value(recipient: AgentEscalationRecipient) -> str | None:
+async def resolve_recipient_value(
+    recipient: AgentEscalationRecipient,
```
ctiliescuuipath: I am not sure if here we want to have other dto (1:1 with AgentEscalationRecipient) just avoid using dtos used at agent.json desearialization cc @valentinabojan ?

ctiliescuuipath (approved):

PR #496, opened 2026-01-29
feat: Integrate UiPath-LangChain-Client and cleanup old clients

cosminacho:
This new client package has 2 main components:

1. uipath-llm-client
- this is the core package on which uipath-langchain-client depends
- it handles URL, headers and authentification for both AgentHub and LLMGateway
- Encapsulates all the auth settings as well as retries, custom exceptions, byo stuff, custom headers for streaming etc into an httx.Client
- it also has passthrough httpx clients for providers like openai, anthropic, gemini (using their clients)
- can be added with: uv add "uipath-llm-client[openai, google, all]" etc. Without specifying extra dependencies it installs only the main httpx client that does the url, headers and auth stuff

2. uipath-langchain-client
- uses the httpx client from uipath-llm-client
- the passthrough implementations are minimal, all passthrough clients use a httpx client to make requests, we can just use the httpx client created above and inject it in the langchain-client for that vendor
- it also has a factory method which does a GET on the discovery endpoint and correctly instantiates the appropiate passthrough client for that model_name and the vendor detected

Summary of the Changes for this package:
- Retrying moved to the uipath-langchain-client
- Request Mixin removed because it’s on the uipath-langchain-client now as UiPathBaseLLMClient
- Streaming fixes added to uipath-langchain-client
- Auth headers, url creation and refresh token with uipath auth moved to uipath-langchain-client
- All clients implement UiPathBaseLLMCLient which has api_config from where we can get vendor_type and api_flavor
- tool_choice=“any” works for any provider if you use bind_tools() which should be used. bind() just adds the “any” parameter to the request body, it doesn’t map it to the correct tool_choice format for that provider
- Retryers are handled in uipath-langchain-client at transport level on the httpx client, so no need to hardcode for each provider
- Factory is now in uipath-langchain-client (and it uses the discovery endpoint)
- Gemini streaming fixes moved 
- Comprehensive tests are in the package, we can remove them from here
- Also the normalized client is in the new client

Before merging this PR I think it's important to make sure that all the relevant components have been moved to the new client, and also that the new changes aren't breaking stuff that's working

Still WIP, but maybe not necessary for this merge:
- merging generation info with response metadata on gemini? (i need to test it to see what's the problem, and add the fix there)
- on Bedrock: Convert file_blocks_to_anthropic_documents: to test if it's really working without it. But in anycase, I think it's better to use UiPathChatAnthropic instead of the BedRock Client. I will test this functionality with both to see the differences.

Some personal recommendations:
- I think for anthropic it's better to use UiPathChatAnthropic(model="etc", provider_type="vertexai"/"bedrock"), instead of doing the Bedrock clients. Why? Because UiPathChatAnthropic has the latest features that anthropic releases on new models, and in it's implementation uses specific http clients for each provider (also maintained by them). So I think it's better to trust the model provider that it keeps the api in sync, and not AWS which hosts a 1000 different models on the same API

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-langchain==1.0.0.dev1004962885",

  # Any version from PR
  "uipath-langchain>=1.0.0.dev1004960000,<1.0.0.dev1004970000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```


src/uipath_langchain/chat/models.py
```diff
@@ -1,482 +0,0 @@
...
-from langchain_openai.chat_models import AzureChatOpenAI
-from pydantic import BaseModel
-from uipath.utils import EndpointManager
-
-from uipath_langchain._utils._request_mixin import UiPathRequestMixin
-
-logger = logging.getLogger(__name__)
-
-
-class UiPathAzureChatOpenAI(UiPathRequestMixin, AzureChatOpenAI):
```
cristipufu: these are public contracts - you need to make sure we don't break existing integrations

pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
     "jsonpath-ng>=1.7.0",
     "mcp==1.24.0",
     "langchain-mcp-adapters==0.2.1",
+    "uipath-langchain-client[all]>=0.2.9",
```
cristipufu: We need to transfer these packags to the official UiPath pypi feed

samples/bring-your-own-model/main.py
```diff
@@ -1,6 +1,6 @@
 from langchain_core.messages import HumanMessage, SystemMessage
 from langgraph.graph import START, StateGraph, END
-from uipath_langchain.chat import UiPathChatOpenAI
+from uipath_langchain_client.clients.openai import UiPathChatOpenAI
```
cristipufu: keep the namespaces, re-export

pyproject.toml
cosminacho: done

pyproject.toml
ionut-mihalache-uipath: I think you should increase the major version of uipath-langchain-python as it is not backwards compatible. Some methods used in uipath-agents-python are removed.

PR #634, opened 2026-02-26
feat: Add uipath_langchain_client, refactor code to use the new client

cosminacho:
This new client has 2 main components:

1. uipath-llm-client
- this is the core package on which uipath-langchain-client depends
- it handles URL, headers and authentification for both AgentHub and LLMGateway
- Encapsulates all the auth settings as well as retries, custom exceptions, byo stuff, custom headers for streaming etc into an httx.Client
- it also has passthrough httpx clients for providers like openai, anthropic, gemini (using their clients)
- can be added with: uv add "uipath-llm-client[openai, google, all]" etc. Without specifying extra dependencies it installs only the main httpx client that does the url, headers and auth stuff

2. uipath-langchain-client
- uses the httpx client from uipath-llm-client
- the passthrough implementations are minimal, all passthrough clients use a httpx client to make requests, we can just use the httpx client created above and inject it in the langchain-client for that vendor
- it also has a factory method which does a GET on the discovery endpoint and correctly instantiates the appropiate passthrough client for that model_name and the vendor detected

Summary of the Changes for this package:
- Retrying moved to the uipath-langchain-client
- Request Mixin removed because it’s on the uipath-langchain-client now as UiPathBaseLLMClient
- Streaming fixes added to uipath-langchain-client
- Auth headers, url creation and refresh token with uipath auth moved to uipath-langchain-client
- All clients implement UiPathBaseChatModel which has api_config from where we can get vendor_type and api_flavor
- Retryers are handled in uipath-langchain-client at transport level on the httpx client, so no need to hardcode for each provider
- Factory is now in uipath-langchain-client (and it uses the discovery endpoint)
- Gemini streaming fixes moved 
- Comprehensive tests are in the package, we can remove them from here
- Also the normalized client is in the new client
- merging generation info with response metadata in the client

Before merging this PR I think it's important to make sure that all the relevant components have been moved to the new client, and also that the new changes aren't breaking stuff that's working

Not tested yet:
- on Bedrock: Convert file_blocks_to_anthropic_documents

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-langchain==0.10.0.dev1006344356",

  # Any version from PR
  "uipath-langchain>=0.10.0.dev1006340000,<0.10.0.dev1006350000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath-langchain>=0.10.0.dev1006340000,<0.10.0.dev1006350000",
]
```


andrei-rusu (approved):

radu-mocanu (changes requested):
we are trying to decouple uipath-langchain from uipath library. the uipath-llm-client will need to use uipath-platform here https://github.com/UiPath/uipath-llm-client-python/blob/main/pyproject.toml#L12

we are moving the relevant auth methods as part of this PR: https://github.com/UiPath/uipath-python/pull/1439

ionmincu (approved):

radu-mocanu (changes requested):

testcases/multimodal-invoke/src/main.py
```diff
@@ -1,20 +1,101 @@
...
+    ),
+    # VendorType.VERTEXAI (Google family) -> UiPathChatGoogleGenerativeAI
+    ModelTestConfig(
+        label="VertexAI (Google) - UiPathChatGoogleGenerativeAI",
+        model_name="gemini-3-pro-preview",
+    ),
+    # Direct instantiation -> UiPathChatAnthropicBedrock
+    ModelTestConfig(
+        label="Bedrock (Anthropic, direct) - UiPathChatAnthropicBedrock",
+        model_name="anthropic.claude-sonnet-4-5-20250929-v1:0",
```
ionut-mihalache-uipath, line 86: Can you add the Anthropic messages API flavor in the list of API flavors and explicitly specify it here?

testcases/multimodal-invoke/src/main.py
```diff
@@ -1,20 +1,101 @@
...
+        max_tokens=200,
+    )
+
+
+# Each entry targets a different class returned by get_chat_model_factory
+MODELS_TO_TEST: list[ModelTestConfig] = [
+    # VendorType.OPENAI (UiPath-owned) -> UiPathAzureChatOpenAI
+    ModelTestConfig(
+        label="OpenAI (Azure) - UiPathAzureChatOpenAI",
+        model_name="gpt-5.2-2025-12-11",
```
ionut-mihalache-uipath, line 64: Is the default API flavor chat completion? The old behavior was a default on responses API, we should maintain the old behavior.

cosminacho: AzureChatOpenAI and ChatOpenAI classes default to chat-completions. There are multiple ways in which you can activate responses API:
- set the parameters "use_reponses_api": True
- specify the parameter "reasoning"
- the request body has the responses api format
So these 2 classes switch dynamically between chat-completions and responses, based on the input. The same is true for UiPathChatOpenAI and UiPathAzureChatOpenAI. But the default on this class is "chat-completions" and I think we shouldn't override the default behaviour of the parent class.

What we can do is modify the factory function "get_chat_model" and set responses there as default such that when the client class is instantiated, it has the parameter "use_responses_api" set to true.

But i still think there might be some problems with BYOM, because the OpenAI-compliant format is the "chat-completion" (afaik).

testcases/multimodal-invoke/src/main.py
cosminacho: Is this a new api flavour in LLMGW? I'm not aware. ChatAnthropicBedrock works only with "invoke" api flavour on bedrock.

testcases/multimodal-invoke/src/main.py
ionut-mihalache-uipath: I understand that the default for AzureChatOpenAI is chat completion, but what I meant is: if the discovery API from LLMGW does not return an API flavor (meaning both api flavors are supported), we must explicitly set use_reponses_api to True in order to not change how the clients were instantiated until now (before this PR). Same for Anthropic: if no API flavor is provided, use UiPathChatBedrockConverse.

Regarding BYO, it was agreed with the LLMGW team that if the API Flavor is not sent by them (meaning both API Flavors are supported) to use the same defaults as for UiPath models. Changing the defaults might change the behavior of deployed agents.

testcases/multimodal-invoke/src/main.py
ionut-mihalache-uipath: You are right, LLMGW should add the new API flavor first.

ionut-mihalache-uipath (approved):

ionut-mihalache-uipath (approved):

pyrightconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "include": ["src"],
```
radu-mocanu: let s keep this PR scope focused. we can include pyright in a future PR

radu-mocanu (approved):

PR #703, opened 2026-03-17
feat: confirmToolCall support — startToolCall with requireConfirmation + ToolRunnableCallable [JAR-9208]

JoshParkSJ:
### Changes
- Moves tool confirmation from the old interrupt-based flow to `confirmToolCall` on `startToolCall`:
1. **Graph build** — LangGraph's wrapping discards the original `BaseTool` reference; the typed subclass preserves `.tool` so the runtime can inspect it at init time (so we know which tools need confirmation before hitting the tool node and can pass require confirmation on `startToolCall`)
2. **Runtime init** — `_get_tool_confirmation_info` walks the compiled graph once, finds `ConversationalToolRunnableCallable` nodes, and builds a `{tool_name: input_schema}` lookup. This is needed because coded agents (`create_agent`) export a compiled graph as the only artifact — there's no side channel to pass confirmation metadata from the build step to the runtime.
3. **Streaming** — mapper checks the lookup on every `AIMessage` tool call, emits `startToolCall` with `requireConfirmation: true` and `inputSchema`.

- fixes `AIMessageChunk` self-accumulation bug that doubled tool names on the first streamed chunk (only triggered with OpenAI, not Claude); `startToolCall` now always emitted for confirmation tools
- rip-and-replace as we see zero production usage confirmed via App Insights telemetry (90 days)

### Companion PRs

- [AgentInterfaces#899](https://github.com/UiPath/AgentInterfaces/pull/899) — CAS frontend
- [uipath-python#1558](https://github.com/UiPath/uipath-python/pull/1558) — confirmToolCall model + bridge
- [uipath-runtime-python#104](https://github.com/UiPath/uipath-runtime-python/pull/104) — remove emit_interrupt_event
- [uipath-agents-python#420](https://github.com/UiPath/uipath-agents-python/pull/420) — skip HITL during eval

https://uipath.atlassian.net/browse/JAR-9208

### DIFF
--- a/src/uipath_langchain/agent/tools/tool_node.py
+++ b/src/uipath_langchain/agent/tools/tool_node.py
@@ -291,7 +291,7 @@ def _wrap_tool_error_handling(
 
 
 class ConversationalToolRunnableCallable(RunnableCallable):
-    """Preserves a reference to the underlying BaseTool for conversational HITL confirmation."""
+    """Preserves a reference to the underlying BaseTool for conversational tool confirmation."""
 
     def __init__(self, *, func: Any, afunc: Any, name: str, tool: BaseTool):
         super().__init__(func=func, afunc=afunc, name=name)
--- a/src/uipath_langchain/agent/tools/tool_node.py
+++ b/src/uipath_langchain/agent/tools/tool_node.py
@@ -22,7 +22,10 @@ from uipath_langchain.agent.react.utils import (
     extract_current_tool_call_index,
     find_latest_ai_message,
 )
-from uipath_langchain.chat.hitl import request_conversational_tool_confirmation
+from uipath_langchain.chat.hitl import (
+    REQUIRE_CONVERSATIONAL_CONFIRMATION,
+    request_conversational_tool_confirmation,
+)
 
 # the type safety can be improved with generics
 ToolWrapperReturnType = dict[str, Any] | Command[Any] | None
@@ -274,7 +277,32 @@ def _wrap_tool_error_handling(
                 raise
             return result
 
+    tool = getattr(tool_node, "tool", None)
+
+    # Preserve tool ref so the runtime can discover which tools need confirmation
+    # (see runtime.py _get_tool_confirmation_info)
+    metadata = getattr(tool, "metadata", None) or {}
+    if isinstance(tool, BaseTool) and metadata.get(REQUIRE_CONVERSATIONAL_CONFIRMATION):
+        return RunnableCallableWithTool(
+            func=_func, afunc=_afunc, name=tool_name, tool=tool
+        )
+
     return RunnableCallable(func=_func, afunc=_afunc, name=tool_name)
+
+
+class RunnableCallableWithTool(RunnableCallable):
+    """A RunnableCallable that preserves a reference to its underlying BaseTool."""
+
+    def __init__(
+        self,
+        *,
+        func: Callable[..., Any] | None,
+        afunc: Callable[..., Awaitable[Any]] | None,
+        name: str,
+        tool: BaseTool,
+    ):
+        super().__init__(func=func, afunc=afunc, name=name)
+        self.tool = tool
 
 
 class ToolWrapperMixin:
--- a/src/uipath_langchain/agent/tools/tool_node.py
+++ b/src/uipath_langchain/agent/tools/tool_node.py
@@ -22,7 +22,10 @@ from uipath_langchain.agent.react.utils import (
     extract_current_tool_call_index,
     find_latest_ai_message,
 )
-from uipath_langchain.chat.hitl import request_conversational_tool_confirmation
+from uipath_langchain.chat.hitl import (
+    REQUIRE_CONVERSATIONAL_CONFIRMATION,
+    request_conversational_tool_confirmation,
+)
 
 # the type safety can be improved with generics
 ToolWrapperReturnType = dict[str, Any] | Command[Any] | None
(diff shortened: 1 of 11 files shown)


maxduu (approved):

src/uipath_langchain/runtime/runtime.py
```diff
@@ -490,17 +488,36 @@ def _detect_middleware_nodes(self) -> set[str]:
...
 
-    def _get_tool_names_requiring_confirmation(self) -> set[str]:
-        names: set[str] = set()
+    def _get_tool_confirmation_info(self) -> dict[str, Any]:
+        """Build {tool_name: input_schema} for tools requiring confirmation.
+
+        Walks compiled graph nodes once at runtime init. This is needed because coded agents
+        (create_agent) export a compiled graph as the only artifact — there's no side channel
+        to pass confirmation metadata from the build step to the runtime.
+        """
```
JoshParkSJ, line 499: not a problem for low code agents, but might as well unify how we get tools requiring confirmation

tests/runtime/test_chat_message_mapper.py
```diff
@@ -1996,26 +1996,38 @@ async def test_pii_masked_response_full_flow(self):
...
+    """Return True if the installed uipath-core exposes the require_confirmation field."""
+    try:
+        from uipath.core.chat.tool import (
+            UiPathConversationToolCallStartEvent,
+        )
+
+        return hasattr(UiPathConversationToolCallStartEvent, "require_confirmation")
+    except Exception:
+        return False
+
```
maxduu: Let's remove this

src/uipath_langchain/runtime/messages.py
```diff
@@ -340,7 +340,13 @@ async def map_ai_message_chunk_to_events(
...
                     case "tool_call_chunk":
                         # Accumulate the message chunk. Note that we assume no interweaving of AIMessage and AIMessageChunks for a given message.
-                        if isinstance(self.current_message, AIMessageChunk):
+                        # Skip the first chunk — it's already assigned as current_message above,
+                        # so accumulating it with itself would duplicate fields via string concat
+                        # (e.g. tool name "search_web" becomes "search_websearch_web").
+                        if (
+                            isinstance(self.current_message, AIMessageChunk)
+                            and self.current_message is not message
+                        ):
```
maxduu: Is this what causes the duplicate tool name issue? I remember @shannonsuhendra  wrote that this issue wasn't [happening](https://uipath.atlassian.net/browse/JAR-9382) anymore.
If this change is risky maybe we can move to a different PR?

src/uipath_langchain/agent/tools/tool_node.py
```diff
@@ -274,9 +277,27 @@ async def _afunc(state: AgentGraphState) -> OutputType:
...
     return RunnableCallable(func=_func, afunc=_afunc, name=tool_name)
 
 
+class ConversationalToolRunnableCallable(RunnableCallable):
+    """Preserves a reference to the underlying BaseTool for conversational tool confirmation."""
+
+    def __init__(self, *, func: Any, afunc: Any, name: str, tool: BaseTool):
+        super().__init__(func=func, afunc=afunc, name=name)
+        self.tool = tool
+
```
maxduu, line 306: This part seems a little sketchy to me in terms of typing. It would be nice if the runtime.py code has the `tool` field typed, or at least can do an `if isinstance ...` to get the right class typing.
Some questions:
- Any way we can add the fields needed by the `tool` to some metadata within `RunnableCallable`?
- At the very least we don't need to do `any` for `func`, `afunc` in `ConversationalToolRunnableCallable`? But if we could do the first bullet it might be even better

Would be nice to get a look from @cristipufu when approach is finalized as well.

src/uipath_langchain/runtime/messages.py
JoshParkSJ: It happens only for GPT models. You can see it happen consistently in staging rn

Screenshot 2026-04-22 at 3 17 57 PM

JoshParkSJ: updated the comment so it's clear this is for gpt only bug

src/uipath_langchain/agent/tools/tool_node.py
JoshParkSJ: Unfortunately `RunnableCallable` is a langgraph class, not ours to touch. Fixed `func`/`afunc` and `if isintance ...`

If you think `ConversationalToolRunnableCallable` is just ugly/excessive then we can force assign `node.tool` like this. Not sure if that's better though

```
    node = RunnableCallable(func=_func, afunc=_afunc, name=tool_name)

    tool = getattr(tool_node, "tool", None)
    if isinstance(tool, BaseTool):
        node.tool = tool  # type: ignore[attr-defined]

    return node
```

src/uipath_langchain/runtime/messages.py
maxduu: Nice thanks for fixing!

src/uipath_langchain/agent/tools/tool_node.py
maxduu: I think having a separate class is fine - maybe some small suggestions:
- Maybe `RunnableCallableWithTool` in case autonomous ever needs it too down the line? Also its a bit more descriptive since nothing about adding the `tool` is directly implying something "conversational".
- Maybe we could we do an `if isinstance` of the class when its needed, to get proper typing? But if the rest of codebase there doesn't do that, might not be a big deal

maxduu (approved):
Just a reply

PR #735, opened 2026-03-27
feat: resolve ArgumentEmail and ArgumentGroupName recipients at runtime

dushyant-uipath:
## Summary

Updates `resolve_recipient_value()` in `escalation_tool.py` to handle the two new argument-driven recipient types added in UiPath/uipath-python PR #1521. At runtime, each type's `argument_path` is looked up in the agent's input kwargs using dot-notation (`safe_get_nested`) to produce a concrete `TaskRecipient`.

Also fixes `create_mcp_tools_from_agent` — the function was incorrectly aliased to `create_mcp_tools_and_clients` (which takes `list[AgentMcpResourceConfig]`), but callers pass a `LowCodeAgentDefinition`. Replaced with a proper wrapper that extracts MCP resources from the agent definition first.

## Changed files

| File | Change |
|---|---|
| `escalation_tool.py` | `resolve_recipient_value` accepts optional `input_args`; `ArgumentEmailRecipient` and `ArgumentGroupNameRecipient` handling added; call site passes filtered agent input (internal graph state fields excluded) |
| `escalate_action.py` | Guardrails path also passes `input_args` to `resolve_recipient_value` — previously always raised `ValueError` for argument-based recipients |
| `mcp/__init__.py` + `mcp_tool.py` | `create_mcp_tools_from_agent` definition moved to `mcp_tool.py` with correct `list[BaseTool]` return type; re-exported from `__init__.py` |
| `pyproject.toml` | `uipath>=2.10.49` (requires the companion PR to be merged) |

### The only runtime failure path
In `resolve_recipient_value()`:
```python
value = safe_get_nested(input_args or {}, recipient.argument_path)
if value is None:
    raise ValueError(
        f"Argument '{recipient.argument_path}' has no value in agent input."
    )
```
This raises when:
1. The agent is invoked without the expected input argument (key missing)
2. The argument is present but explicitly `None`

This is intentional — a misconfigured escalation should fail loudly at task creation time, not silently create a task with no assignee.

**There is no feature flag in the Python layer.** The flag (`EnableEscalationArgumentAssignee`) lives in the frontend and Agents backend only. If the frontend sends an `ArgumentEmail`/`ArgumentGroupName` recipient (flag is on), the Python side always processes it. The frontend strips these types from the payload when the flag is off.

### End-to-end local test (hacked-coded, 2026-04-15)

Agent configured with `ArgumentEmail` recipient (`argument_path="assigneeEmail"`), run against alpha:

```
uv run uipath run agent.json '{"assigneeEmail": "dushyant.pathak@uipath.com", "taskDescription": "Test dynamic assignee"}'
```

Action Center task created with:
```
recipient:
  type: TaskRecipientType.EMAIL
  value: dushyant.pathak@uipath.com   ← resolved from input arg at runtime ✓
title: Review Required
data.taskDetails: Test dynamic assignee
```
`✓ Successful execution.`

### Other paths verified
- `create_mcp_tools_from_agent` loads without error — agents factory registers correctly ✓
- Existing `AssetRecipient` and `StandardRecipient` resolution paths unaffected ✓
- `input_args=None` default: all existing callers that don't pass `input_args` continue to work ✓
- `escalate_action.py` guardrails path: `input_args` now passed correctly, no longer raises on argument-based recipients ✓

## Related

- Agents repo PR: UiPath/Agents#4835
- Data model PR: UiPath/uipath-python#1521 (**must merge first**)
- Design doc: https://uipath.atlassian.net/wiki/spaces/BPA/pages/90469990611/Dynamic+Assignee

## Test plan

- [x] `ArgumentEmail` recipient end-to-end: agent run with `assigneeEmail` input → Action Center task created with correct assignee ✓
- [ ] Same for `ArgumentGroupName`
- [ ] Run without the argument populated — verify `ValueError` with clear message
- [ ] Verify existing `AssetRecipient` and `StandardRecipient` paths unaffected
- [ ] Verify `create_mcp_tools_from_agent` no longer crashes when agent has MCP resources

### DIFF
--- a/src/uipath_langchain/agent/tools/escalation_tool.py
+++ b/src/uipath_langchain/agent/tools/escalation_tool.py
@@ -11,9 +11,12 @@ from uipath.agent.models.agent import (
     AgentEscalationRecipient,
     AgentEscalationRecipientType,
     AgentEscalationResourceConfig,
+    ArgumentEmailRecipient,
+    ArgumentGroupNameRecipient,
     AssetRecipient,
     StandardRecipient,
 )
+from uipath.agent.utils.text_tokens import safe_get_nested
 from uipath.eval.mocks import mockable
 from uipath.platform import UiPath
 from uipath.platform.action_center.tasks import TaskRecipient, TaskRecipientType
@@ -46,6 +49,7 @@ class EscalationAction(str, Enum):
 
 async def resolve_recipient_value(
     recipient: AgentEscalationRecipient,
+    input_args: dict[str, Any] | None = None,
 ) -> TaskRecipient | None:
     """Resolve recipient value based on recipient type."""
     if isinstance(recipient, AssetRecipient):
@@ -56,6 +60,26 @@ async def resolve_recipient_value(
         elif recipient.type == AgentEscalationRecipientType.ASSET_GROUP_NAME:
             type = TaskRecipientType.GROUP_NAME
         return TaskRecipient(value=value, type=type, displayName=value)
+
+    if isinstance(recipient, ArgumentEmailRecipient):
+        value = safe_get_nested(input_args or {}, recipient.argument_path)
+        if value is None:
+            raise ValueError(
+                f"Argument '{recipient.argument_path}' has no value in agent input."
+            )
+        return TaskRecipient(
+            value=value, type=TaskRecipientType.EMAIL, displayName=value
+        )
+
+    if isinstance(recipient, ArgumentGroupNameRecipient):
+        value = safe_get_nested(input_args or {}, recipient.argument_path)
+        if value is None:
+            raise ValueError(
+                f"Argument '{recipient.argument_path}' has no value in agent input."
+            )
+        return TaskRecipient(
+            value=value, type=TaskRecipientType.GROUP_NAME, displayName=value
+        )
 
     if isinstance(recipient, StandardRecipient):
         type = TaskRecipientType(recipient.type)
@@ -156,8 +180,11 @@ def create_escalation_tool(
     _bts_context: dict[str, Any] = {}
 
     async def escalation_tool_fn(**kwargs: Any) -> dict[str, Any]:
+        agent_input: dict[str, Any] = (
+            tool.metadata.get("agent_input") if tool.metadata else None
+        ) or {}
         recipient: TaskRecipient | None = (
-            await resolve_recipient_value(channel.recipients[0])
+            await resolve_recipient_value(channel.recipients[0], input_args=agent_input)
             if channel.recipients
             else None
         )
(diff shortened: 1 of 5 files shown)


radugheo:
The method `resolve_recipient_value()` is also called inside `EscalateAction._create_task_node()` in `escalate_action.py`, without input_args. For the new argument-based recipient types, this will always raise a ValueError since input_args defaults to `{}`.

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -249,11 +284,13 @@ async def escalation_wrapper(
...
             raise RuntimeError("Tool metadata is required for task_title resolution")
 
+        state_dict = sanitize_dict_for_serialization(dict(state))
         tool.metadata["task_title"] = resolve_task_title(
             channel.task_title,
-            sanitize_dict_for_serialization(dict(state)),
+            state_dict,
             default_title="Escalation Task",
         )
+        tool.metadata["agent_input"] = state_dict
```
radugheo: `state_dict` is the serialized graph state, which contains keys like `messages` and `inner_state`. 
`safe_get_nested` looks up recipient.argument_path against these keys, so it will return `None` unless the path happens to start with `messages` or `inner_state`.

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -57,6 +69,26 @@ async def resolve_recipient_value(
             type = TaskRecipientType.GROUP_NAME
         return TaskRecipient(value=value, type=type, displayName=value)
 
+    if ArgumentEmailRecipient is not None and isinstance(recipient, ArgumentEmailRecipient):
+        value = safe_get_nested(input_args or {}, recipient.argument_path)
+        if not value:
```
radugheo: Nit: `if not value` would also reject "", 0, False
I think `if value is None` would be more precise here.

src/uipath_langchain/agent/tools/mcp/__init__.py
```diff
@@ -8,11 +10,38 @@
...
 
+if TYPE_CHECKING:
+    from uipath.agent.models.agent import AgentMcpResourceConfig, LowCodeAgentDefinition
+
+
+async def create_mcp_tools_from_agent(
+    agent: "LowCodeAgentDefinition",
+    session_info_factory: SessionInfoFactory | None = None,
+    terminate_on_close: bool = True,
+) -> tuple[list[Any], list[McpClient]]:
```
radugheo: `list[Any]` could be `list[BaseTool]` to match what `create_mcp_tools_and_clients` actually returns

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -37,6 +45,9 @@
 )
 
 
+logger = logging.getLogger(__name__)
```
radugheo: Looks like logger isn't used anywhere, we can remove it

src/uipath_langchain/agent/tools/mcp/__init__.py
```diff
@@ -8,11 +10,38 @@
 )
 from .streamable_http import SessionInfo
 
+if TYPE_CHECKING:
+    from uipath.agent.models.agent import AgentMcpResourceConfig, LowCodeAgentDefinition
+
+
+async def create_mcp_tools_from_agent(
```
radugheo: Would it make more sense to put this in `mcp_tool.py` alongside the other factory functions?

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -14,6 +15,13 @@
     AssetRecipient,
     StandardRecipient,
 )
+
+try:
+    from uipath.agent.models.agent import ArgumentEmailRecipient, ArgumentGroupNameRecipient  # type: ignore[attr-defined]
+except ImportError:  # uipath < 2.10.49 — argument-driven recipients not yet available
```
radugheo: I think we can bump the `uipath` floor version so that these imports are included, and then do a clean import

radugheo (approved):

PR #761, opened 2026-04-09
feat: add A2A tool for invoking remote agents at runtime

PopescuTudor:
Adds an A2A LangChain tool that enables low-code agents to invoke Remote A2A agents at runtime. When an agent definition includes an A2A resource (`$resourceType: "a2a"`), the tool factory now creates an `A2aTool` that sends messages through the AgentHub A2A proxy.

Two conversation persistence modes are supported:
- **Graph state** (`advancedPersistance: true`): task_id/context_id stored in LangGraph's tools_storage
- **LLM passthrough** (default): IDs included in tool schema so the LLM tracks conversation continuity

Uses the `a2a-sdk` Python client for A2A protocol communication, with lazy auth via UiPath Bearer token.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-langchain==0.9.25.dev1007614040",

  # Any version from PR
  "uipath-langchain>=0.9.25.dev1007610000,<0.9.25.dev1007620000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath-langchain>=0.9.25.dev1007610000,<0.9.25.dev1007620000",
]
```

### DIFF
--- a/src/uipath_langchain/agent/tools/a2a/a2a_tool.py
+++ b/src/uipath_langchain/agent/tools/a2a/a2a_tool.py
@@ -1,0 +1,371 @@
+"""A2A singleton tool — one tool per remote agent.
+
+Each tool maintains conversation context (task_id/context_id) across calls
+using deterministic persistence via LangGraph graph state (tools_storage).
+
+Authentication uses the UiPath SDK Bearer token, resolved lazily on first call.
+Client lifecycle is managed by the caller via ``A2aClient.dispose()`` or the
+``open_a2a_tools`` async context manager.
+"""
+
+import asyncio
+import json
+from contextlib import AsyncExitStack, asynccontextmanager
+from logging import getLogger
+from typing import AsyncGenerator
+from uuid import uuid4
+
+import httpx
+from a2a.client import Client
+from a2a.types import (
+    AgentCard,
+    Message,
+    Part,
+    Role,
+    Task,
+    TaskArtifactUpdateEvent,
+    TaskState,
+    TextPart,
+)
+from langchain_core.messages import ToolCall, ToolMessage
+from langchain_core.tools import BaseTool
+from langgraph.types import Command
+from pydantic import BaseModel, Field
+from uipath._utils._ssl_context import get_httpx_client_kwargs
+from uipath.agent.models.agent import AgentA2aResourceConfig
+
+from uipath_langchain.agent.react.types import AgentGraphState
+from uipath_langchain.agent.tools.base_uipath_structured_tool import (
+    BaseUiPathStructuredTool,
+)
+from uipath_langchain.agent.tools.tool_node import (
+    ToolWrapperMixin,
+    ToolWrapperReturnType,
+)
+from uipath_langchain.agent.tools.utils import sanitize_tool_name
+
+logger = getLogger(__name__)
+
+
+class A2aToolInput(BaseModel):
+    """Input schema for A2A agent tool."""
+
+    message: str = Field(description="The message to send to the remote agent.")
+
+
+class A2aStructuredToolWithWrapper(BaseUiPathStructuredTool, ToolWrapperMixin):
+    pass
+
+
+class A2aClient:
+    """Wraps an A2A client and its underlying httpx.AsyncClient for lifecycle management.
+
+    The A2A client is initialized lazily on first ``get()`` call to avoid blocking
+    tool creation. The caller must call ``dispose()`` to close the HTTP connection
+    pool when done.
+    """
+
+    def __init__(self, agent_card: AgentCard) -> None:
+        self._agent_card = agent_card
+        self._lock = asyncio.Lock()
+        self._client: Client | None = None
+        self._http_client: httpx.AsyncClient | None = None
+
+    async def get(self) -> Client:
+        """Get (or lazily create) the A2A client."""
+        if self._client is None:
+            async with self._lock:
+                if self._client is None:
+                    from a2a.client import ClientConfig, ClientFactory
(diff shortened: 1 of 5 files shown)


src/uipath_langchain/agent/tools/a2a/a2a_tool.py
```diff
@@ -0,0 +1,317 @@
...
+                parts.append(f"Skill: {skill_desc}")
+    return " | ".join(parts) if parts else f"Remote A2A agent at {card.url}"
+
+
+def _resolve_a2a_url(config: AgentA2aResourceConfig) -> str:
+    """Resolve the A2A endpoint URL from config."""
+    a2a_url = getattr(config, "a2a_url", None)
+    if a2a_url:
+        return a2a_url
+    return config.agent_card_url.replace("/.well-known/agent-card.json", "")
```
radugheo: after we merge [this PR](https://github.com/UiPath/uipath-python/pull/1554), let's change this method to use the url from the `cached_agent_card` field:
```
def _resolve_a2a_url(config: AgentA2aResourceConfig) -> str:
    """Resolve the A2A endpoint URL from config."""
    if config.cached_agent_card and "url" in config.cached_agent_card:
        return config.cached_agent_card["url"]
    return ""
```

src/uipath_langchain/agent/tools/a2a/a2a_tool.py
```diff
@@ -0,0 +1,317 @@
...
+
+    Conversation context (task_id/context_id) is persisted deterministically
+    in LangGraph's graph state via tools_storage, ensuring reliable
+    multi-turn conversations with the remote agent.
+    """
+    if config.cached_agent_card:
+        agent_card = AgentCard(**config.cached_agent_card)
+    else:
+        agent_card = AgentCard(
+            url=config.agent_card_url,
```
radugheo: we should update the url here too

src/uipath_langchain/agent/tools/a2a/a2a_tool.py
```diff
@@ -0,0 +1,317 @@
...
+    Returns:
+        List of BaseTool instances, one per enabled A2A resource.
+    """
+    tools: list[BaseTool] = []
+
+    for resource in resources:
+        if resource.is_enabled is False:
+            logger.info("Skipping disabled A2A resource '%s'", resource.name)
+            continue
+        if resource.is_active is False:
```
radugheo: this property won't be used anymore, so we can remove the `is_active` check

src/uipath_langchain/agent/tools/a2a/a2a_tool.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+def _resolve_a2a_url(config: AgentA2aResourceConfig) -> str:
+    """Resolve the A2A endpoint URL from the cached agent card."""
+    if config.cached_agent_card and "url" in config.cached_agent_card:
+        return config.cached_agent_card["url"]
+    return ""
+
+
+async def create_a2a_agent_tools(
```
radugheo: This method shouldn't be async

src/uipath_langchain/agent/tools/a2a/a2a_tool.py
```diff
@@ -0,0 +1,312 @@
...
+    raw_name = agent_card.name or config.name
+    tool_name = sanitize_tool_name(raw_name)
+    tool_description = _build_description(agent_card)
+    a2a_url = _resolve_a2a_url(config)
+
+    _lock = asyncio.Lock()
+    _client: Client | None = None
+    _http_client: httpx.AsyncClient | None = None
+
+    async def _ensure_client() -> Client:
```
radugheo: The `httpx.AsyncClient` is never closed. We could add a context manager like `open_mcp_tools`  (from the mcp tool) to handle cleanup.

src/uipath_langchain/agent/tools/a2a/a2a_tool.py
```diff
@@ -0,0 +1,312 @@
...
+            if skill_desc:
+                parts.append(f"Skill: {skill_desc}")
+    return " | ".join(parts) if parts else f"Remote A2A agent at {card.url}"
+
+
+def _resolve_a2a_url(config: AgentA2aResourceConfig) -> str:
+    """Resolve the A2A endpoint URL from the cached agent card."""
+    if config.cached_agent_card and "url" in config.cached_agent_card:
+        return config.cached_agent_card["url"]
+    return ""
```
radugheo: Maybe we should raise an error here instead of passing with an empty string

src/uipath_langchain/agent/tools/a2a/a2a_tool.py
PopescuTudor: Fixed. Restructured following the MCP pattern:

- Added `A2aClient` class wrapping the httpx client with a `dispose()` method
- `create_a2a_tools_and_clients()` returns `(tools, clients)` like `create_mcp_tools_and_clients`
- Added `open_a2a_tools()` async context manager for convenience
- Removed A2A handling from `tool_factory.py` - now wired in `build_agent_graph` with the clients added to the disposables list, so the runtime closes them on teardown

Depends on companion PR in uipath-agents-python that wires `create_a2a_tools_and_clients` into `build_agent_graph` alongside MCP clients: https://github.com/UiPath/uipath-agents-python/pull/424. Must be merged after this one (and after bumping the `uipath-langchain` pin there).

radugheo (approved):

PR #778, opened 2026-04-17
feat: add ff

ionut-mihalache-uipath:
Add FF

### DIFF
--- a/src/uipath_langchain/chat/chat_model_factory.py
+++ b/src/uipath_langchain/chat/chat_model_factory.py
@@ -1,39 +1,122 @@
-from uipath_langchain_client.base_client import UiPathBaseChatModel
-from uipath_langchain_client.factory import get_chat_model as get_chat_model_factory
-from uipath_langchain_client.settings import (
-    ApiFlavor,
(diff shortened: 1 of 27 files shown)


src/uipath_langchain/chat/chat_model_factory.py
```diff
@@ -1,39 +1,122 @@
...
-from uipath_langchain_client.factory import get_chat_model as get_chat_model_factory
-from uipath_langchain_client.settings import (
-    ApiFlavor,
-    RoutingMode,
-    UiPathBaseSettings,
-    VendorType,
-    get_default_client_settings,
-)
-
-_PREFERRED_FLAVOR_BY_VENDOR: dict[str, ApiFlavor] = {
```
cosminacho, line 11: Should we keep those defaults?

cosminacho (approved):

src/uipath_langchain/chat/chat_model_factory.py
ionut-mihalache-uipath: discussed offline - they are preserved

PR #793, opened 2026-04-21
feat: add Agent Episodic Memory support (recall node)

mjnovice:
## Summary

- **Memory recall graph node**: When `MemoryConfig` is passed to `create_agent()`, a `MEMORY_RECALL` node is inserted before `INIT` that queries the UiPath Memory service and stores the `systemPromptInjection` in graph state. The INIT node reads it and appends to the system prompt.
- **New types**: `MemoryConfig`, `AgentGraphNode.MEMORY_RECALL`, `InnerAgentGraphState.memory_injection`
- **Dependency**: bumps `uipath-platform >= 0.1.31` for `MemoryService` types (depends on UiPath/uipath-python#1467)

> **Note:** Escalation memory is **not** included in this PR. Only episodic memory recall is being changed.

### Graph with memory enabled
```
START → MEMORY_RECALL → INIT → AGENT ↔ TOOLS → TERMINATE → END
```

### Files changed
- `types.py` — `MemoryConfig` dataclass, `memory_injection` field on inner state, `MEMORY_RECALL` node enum
- `memory_node.py` — New: memory recall graph node
- `agent.py` — `create_agent()` accepts optional `memory` parameter
- `init_node.py` — Reads `memory_injection` from state and appends to system prompt
- `__init__.py` — Exports `MemoryConfig`

## Relationship with UiPath/uipath-agents-python#374

This PR and UiPath/uipath-agents-python#374 together implement the full Agent Episodic Memory feature across both repos:

| Concern | This PR (`uipath-langchain-python`) | UiPath/uipath-agents-python#374 |
|---|---|---|
| **Memory recall graph node** | Implements `MEMORY_RECALL` node, `MemoryConfig`, and INIT-node injection | Resolves `MemoryConfig` from agent definition (`isAgentMemoryEnabled` + `memorySpaceId`) and passes it to `create_agent()` |
| **Memory module** | `memory_node.py` — builds `MemorySearchRequest`, calls `sdk.memory.search_async()` | `memory.py` — contains `build_memory_search_request()` and `retrieve_memory_injection()` helpers (to be replaced by the graph node from this PR) |
| **Factory wiring** | Exposes `MemoryConfig` for callers to pass | `_get_memory_config()` extracts memory settings from `AgentDefinition` and constructs `MemoryConfig` |

**Merge order**: This PR should merge first (or concurrently with UiPath/uipath-python#1467). Then UiPath/uipath-agents-python#374 can be updated to:
1. Remove the runtime-level `_populate_memory_context()` mutable-dict approach
2. Instead, pass `MemoryConfig` to `create_agent(memory=...)` in `graph.py`
3. Remove `memory_context` threading through the message factory

Both PRs depend on UiPath/uipath-python#1467 (`uipath-platform >= 0.1.31`) for the `MemoryService` types.

## Test plan

- [x] 20 new unit tests (memory node, escalation cache, escalation ingest)
- [x] Existing `test_create_agent.py` tests pass (graph structure unchanged when `memory=None`)
- [x] Full suite: 1544 passed (2 pre-existing failures from missing optional deps)
- [ ] E2E test with real memory space (requires UiPath/uipath-python#1467 merged)

E2E flow tested with https://github.com/UiPath/Agents/pull/5082/changes enabled
image

Ingestion
image

### DIFF
--- a/src/uipath_langchain/_utils/_otel.py
+++ b/src/uipath_langchain/_utils/_otel.py
@@ -1,0 +1,15 @@
+"""OpenTelemetry utilities for span and trace context."""
+
+from typing import Any
+
+
+def set_span_attribute(name: str, value: Any) -> None:
+    """Set an attribute on the current OTel span (no-op if unavailable)."""
+    try:
+        from opentelemetry import trace
+
+        span = trace.get_current_span()
+        if span.is_recording():
+            span.set_attribute(name, value)
+    except ImportError:
+        pass
--- a/src/uipath_langchain/agent/react/init_node.py
+++ b/src/uipath_langchain/agent/react/init_node.py
@@ -25,6 +25,19 @@ def create_init_node(
             resolved_messages = list(messages(state))
         else:
             resolved_messages = list(messages)
+
+        # Append memory injection from the MEMORY_RECALL node (if present)
+        memory_injection = ""
+        if hasattr(state, "inner_state") and hasattr(
+            state.inner_state, "memory_injection"
+        ):
+            memory_injection = state.inner_state.memory_injection or ""
+        if memory_injection and resolved_messages:
+            first = resolved_messages[0]
+            if isinstance(first, SystemMessage):
+                resolved_messages[0] = SystemMessage(
+                    content=str(first.content) + memory_injection
+                )
         if is_conversational:
             # For conversational agents we need to reorder the messages so that the system message is first, followed by
             # the initial user message. When resuming the conversation, the state will have the entire message history,
--- a/src/uipath_langchain/agent/react/memory_node.py
+++ b/src/uipath_langchain/agent/react/memory_node.py
@@ -1,0 +1,228 @@
+"""Memory recall node for Agent Episodic Memory.
+
+Queries the UiPath Memory service (via LLMOps) for similar past episodes
+and stores the server-generated systemPromptInjection in graph state so
+the INIT node can append it to the system prompt.
+"""
+
+import logging
+from contextlib import contextmanager
+from typing import Any
+
+from pydantic import BaseModel
+from uipath.platform import UiPath
+from uipath.platform.errors import EnrichedException
+from uipath.platform.memory import (
+    FieldSettings,
+    MemorySearchRequest,
+    SearchField,
+    SearchMode,
+    SearchSettings,
+)
+
+from .types import MemoryConfig
+from .utils import extract_input_data_from_state
+
+logger = logging.getLogger(__name__)
+
+
+@contextmanager
+def _noop_context():
+    """No-op context manager when OTel is unavailable."""
+    yield None
+
+
+def create_memory_recall_node(
+    memory_config: MemoryConfig,
+    input_schema: type[BaseModel] | None = None,
+):
+    """Create an async graph node that retrieves memory injection.
+
+    The node queries ``sdk.memory.search_async()`` and writes the
+    ``systemPromptInjection`` string into ``inner_state.memory_injection``.
+    On failure it logs a warning and continues with an empty injection.
+
+    Args:
+        memory_config: Memory configuration with space ID and search settings.
+
+    Returns:
+        An async callable suitable for ``builder.add_node()``.
+    """
+
+    async def memory_recall_node(state: Any) -> dict[str, Any]:
+        input_model = input_schema if input_schema is not None else BaseModel
+        input_arguments = extract_input_data_from_state(state, input_model)
+        if not input_arguments:
+            logger.debug("Memory recall: no user inputs found in state")
+            return {}
+
+        fields = _build_search_fields(
+            input_arguments, field_weights=memory_config.field_weights or None
+        )
+        if not fields:
+            logger.debug(
+                "Memory recall: no search fields after filtering (inputs=%s, weights=%s)",
+                list(input_arguments.keys()),
+                memory_config.field_weights,
+            )
+            return {}
+
+        request = MemorySearchRequest(
+            fields=fields,
+            settings=SearchSettings(
+                threshold=memory_config.threshold,
+                result_count=memory_config.result_count,
+                search_mode=SearchMode.Hybrid,
+            ),
+            definition_system_prompt="",
+        )
+
+        results_count = 0
+        # Wrap the search in OTel spans so "Find previous memories" and
+        # "Apply dynamic few shot" appear in the Execution Trace with
+        # correct timing. The LlmOpsHttpExporter picks these up.
+        injection = ""
+        try:
+            from opentelemetry import trace as otel_trace
+
+            tracer = otel_trace.get_tracer("uipath_langchain.memory")
+        except ImportError:
+            tracer = None
+            otel_trace = None  # type: ignore[assignment]
+
+        # Span attribute keys matching what the LlmOpsHttpExporter and
+        # Studio UI expect. "openinference.span.kind" sets SpanType.
+        lookup_span_ctx = (
+            tracer.start_as_current_span(
+                "Find previous memories",
+                attributes={
+                    "openinference.span.kind": "agentMemoryLookup",
+                    "type": "agentMemoryLookup",
+                    "span_type": "agentMemoryLookup",
+                    "uipath.custom_instrumentation": True,
+                    "memorySpaceName": memory_config.memory_space_name or "",
+                    "memorySpaceId": memory_config.memory_space_id,
+                    "strategy": "DynamicFewShotPrompt",
+                },
+            )
+            if tracer
+            else _noop_context()
+        )
+
+        with lookup_span_ctx as lookup_span:
+            fewshot_span_ctx = (
+                tracer.start_as_current_span(
+                    "Apply dynamic few shot",
+                    attributes={
+                        "openinference.span.kind": "applyDynamicFewShot",
+                        "type": "applyDynamicFewShot",
+                        "span_type": "applyDynamicFewShot",
+                        "uipath.custom_instrumentation": True,
+                        "memorySpaceName": memory_config.memory_space_name or "",
+                        "memorySpaceId": memory_config.memory_space_id,
+                    },
+                )
+                if tracer
+                else _noop_context()
+            )
+
+            with fewshot_span_ctx as fewshot_span:
+                try:
+                    sdk = UiPath()
+                    folder_key = memory_config.folder_key
+                    if not folder_key and memory_config.folder_path:
+                        folder_key = await sdk.folders.retrieve_folder_key_async(
+                            memory_config.folder_path
+                        )
+                    response = await sdk.memory.search_async(
+                        memory_space_id=memory_config.memory_space_id,
+                        request=request,
+                        folder_key=folder_key,
+                    )
+                    injection = response.system_prompt_injection
+                    results_count = len(response.results)
+                    logger.info(
+                        "Memory recall returned %d results for space '%s'",
+                        results_count,
+                        memory_config.memory_space_id,
+                    )
+                    # Set request/response on fewshot span as JSON strings.
+                    # The exporter parses JSON strings back to objects.
+                    # The UI reads "response" to display matched memory items.
+                    if fewshot_span and hasattr(fewshot_span, "set_attribute"):
+                        import json
+
+                        fewshot_span.set_attribute(
+                            "request",
+                            json.dumps(
+                                request.model_dump(by_alias=True, exclude_none=True)
+                            ),
+                        )
+                        fewshot_span.set_attribute(
+                            "response",
+                            json.dumps(
+                                response.model_dump(by_alias=True, exclude_none=True)
+                            ),
+                        )
+                except Exception as e:
+                    if isinstance(e, EnrichedException):
+                        error_detail = (
+                            f"{e} | status={e.status_code} body={e.response_content}"
+                        )
+                    else:
+                        error_detail = repr(e)
+                    logger.warning(
+                        "Memory recall failed for space '%s': %s",
+                        memory_config.memory_space_id,
+                        error_detail,
+                    )
+                    if otel_trace:
+                        if fewshot_span and hasattr(fewshot_span, "set_status"):
+                            fewshot_span.set_status(
+                                otel_trace.StatusCode.ERROR, error_detail
+                            )
+                        if lookup_span and hasattr(lookup_span, "set_status"):
+                            lookup_span.set_status(
+                                otel_trace.StatusCode.ERROR, error_detail
+                            )
+
+            # Set result attributes after search completes
+            if lookup_span and hasattr(lookup_span, "set_attribute"):
+                lookup_span.set_attribute("memoryItemsMatched", results_count)
+                if injection:
+                    lookup_span.set_attribute("result", injection)
+
+        if not injection:
+            return {}
+
+        return {"inner_state": {"memory_injection": injection}}
+
+    return memory_recall_node
+
+
+def _build_search_fields(
+    input_arguments: dict[str, Any],
+    field_weights: dict[str, float] | None = None,
+    field_type: str = "agent-input",
+) -> list[SearchField]:
+    """Convert agent input arguments to SearchField objects.
+
+    The key_path must be prefixed with the field type:
+      keyPath = [fieldType, fieldName]
+    e.g. ["agent-input", "a"] for episodic memory.
+    """
+    fields: list[SearchField] = []
+    for name, value in input_arguments.items():
+        value_str = str(value) if value is not None else ""
+        if not value_str or name.startswith("uipath__"):
+            continue
+        # When field_weights is specified, only include fields with configured weights
+        if field_weights and name not in field_weights:
+            continue
+        settings = FieldSettings()
+        if field_weights and name in field_weights:
+            settings = FieldSettings(weight=field_weights[name])
+        fields.append(
+            SearchField(key_path=[field_type, name], value=value_str, settings=settings)
+        )
+    return fields
--- a/src/uipath_langchain/agent/react/types.py
+++ b/src/uipath_langchain/agent/react/types.py
@@ -3,7 +3,7 @@ from typing import Annotated, Any, Hashable, Literal, Optional
 
 from langchain_core.messages import AnyMessage
 from langgraph.graph.message import add_messages
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
 from uipath.agent.react import END_EXECUTION_TOOL, RAISE_ERROR_TOOL
 from uipath.platform.attachments import Attachment
 
@@ -19,6 +19,7 @@ class InnerAgentGraphState(BaseModel):
     job_attachments: Annotated[dict[str, Attachment], merge_dicts] = {}
     initial_message_count: int | None = None
     tools_storage: Annotated[dict[Hashable, Any], merge_dicts] = {}
+    memory_injection: str = ""
 
 
 class InnerAgentGuardrailsGraphState(InnerAgentGraphState):
@@ -55,6 +56,43 @@ class AgentGraphNode(StrEnum):
     TOOLS = "tools"
     TERMINATE = "terminate"
     GUARDED_TERMINATE = "guarded-terminate"
+    MEMORY_RECALL = "memory_recall"
+
+
+class MemoryConfig(BaseModel):
+    """Configuration for Agent Episodic Memory.
+
+    When passed to ``create_agent()``, a MEMORY_RECALL node is added before
+    INIT that queries the memory service and stores the server-generated
+    systemPromptInjection in ``inner_state.memory_injection``.
+    """
+
+    memory_space_id: str = Field(description="GUID of the memory space to query.")
+    memory_space_name: str = Field(
+        default="", description="Name of the memory space (for tracing)."
+    )
+    folder_key: str | None = Field(
+        default=None, description="Folder key for the memory resource."
+    )
+    folder_path: str | None = Field(
+        default=None,
+        description="Folder path for the memory resource. Resolved to folder_key at runtime if folder_key is not set.",
+    )
+    # Defaults match FE episodic memory settings (agentEditor.ts:324-328)
+    result_count: int = Field(default=3, ge=1, le=10)
+    threshold: float = Field(default=0.0, ge=0.0, le=1.0)
+    field_weights: dict[str, float] = Field(
+        description=(
+            "Per-field search weights. Keys are input field names, values are "
+            "weights between 0.0 and 1.0. At least one field must be specified."
+        ),
+    )
+
+    @model_validator(mode="after")
+    def _validate_field_weights(self) -> "MemoryConfig":
+        if not self.field_weights:
+            raise ValueError("field_weights must contain at least one field")
+        return self
 
 
 class AgentGraphConfig(BaseModel):
--- a/tests/agent/react/test_memory_node.py
+++ b/tests/agent/react/test_memory_node.py
@@ -1,0 +1,151 @@
+"""Tests for memory recall node and memory integration in create_agent."""
+
+from typing import Any
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+from langchain_core.language_models import BaseChatModel
+from langchain_core.messages import SystemMessage
+from langchain_core.runnables.graph import Edge
+from langgraph.graph import StateGraph
+from pydantic import BaseModel, Field
+
+from uipath_langchain.agent.react.agent import create_agent
+from uipath_langchain.agent.react.memory_node import (
+    _build_search_fields,
+    create_memory_recall_node,
+)
+from uipath_langchain.agent.react.types import (
+    AgentGraphNode,
+    MemoryConfig,
+)
+
+
+class _TopicInput(BaseModel):
+    """Reusable test input schema."""
+
+    topic: str = Field(default="")
+
+
+class _TopicLevelInput(BaseModel):
+    """Reusable test input schema with two fields."""
+
+    topic: str = Field(default="")
+    level: str = Field(default="")
+
+
+def _make_state(**user_fields: Any) -> dict[str, Any]:
+    """Create a state dict with standard internal fields + user fields."""
+    return {"messages": [], "inner_state": {}, **user_fields}
+
+
+class TestBuildSearchFields:
+    def test_basic_fields(self) -> None:
+        fields = _build_search_fields({"topic": "python", "level": "advanced"})
+        assert len(fields) == 2
+        key_paths = [f.key_path for f in fields]
+        assert ["agent-input", "topic"] in key_paths
+        assert ["agent-input", "level"] in key_paths
+
+    def test_filters_none_and_uipath_prefix(self) -> None:
+        fields = _build_search_fields(
+            {"topic": "py", "uipath__settings": {}, "empty": None}
+        )
+        assert len(fields) == 1
+        assert fields[0].key_path == ["agent-input", "topic"]
+
+    def test_empty_input(self) -> None:
+        assert _build_search_fields({}) == []
+
+
+class TestMemoryRecallNode:
+    @pytest.mark.asyncio
+    @patch("uipath_langchain.agent.react.memory_node.UiPath")
+    async def test_returns_injection_on_success(
+        self, mock_uipath_cls: MagicMock
+    ) -> None:
+        mock_sdk = MagicMock()
+        mock_uipath_cls.return_value = mock_sdk
+        mock_response = MagicMock()
+        mock_response.results = [MagicMock()]
+        mock_response.system_prompt_injection = "\n\nBased on past runs..."
+        mock_sdk.memory.search_async = AsyncMock(return_value=mock_response)
+
+        config = MemoryConfig(memory_space_id="space-123", field_weights={"topic": 1.0})
+        node = create_memory_recall_node(config, input_schema=_TopicInput)
+
+        result = await node(_make_state(topic="python"))
+        assert result["inner_state"]["memory_injection"] == "\n\nBased on past runs..."
+
+    @pytest.mark.asyncio
+    @patch("uipath_langchain.agent.react.memory_node.UiPath")
+    async def test_returns_empty_on_failure(self, mock_uipath_cls: MagicMock) -> None:
+        mock_sdk = MagicMock()
+        mock_uipath_cls.return_value = mock_sdk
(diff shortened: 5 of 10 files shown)


src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -262,6 +285,13 @@ async def create_escalation_task():
             EscalationAction(outcome_str) if outcome_str else EscalationAction.CONTINUE
         )
 
+        # --- Escalation memory: persist outcome for future recall ---
+        await _ingest_escalation_memory(
+            _memory_space_id,
+            answer=json.dumps(escalation_output),
```
swathiJayav: We need to store Escalation outcome as well. In Temporal, we store a json containing the output/result and outcome. Need to do the same treatment here

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -262,6 +285,13 @@ async def create_escalation_task():
...
+            attributes=json.dumps(serialized_data),
```
swathiJayav: Make sure `serialied_data` has the same shape as here: https://vscode.dev/github/UiPath/Agents/blob/main/backend/Execution.Shared/Tools/EscalationTool/EscalationToolExecutor.cs#L503

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -333,3 +363,98 @@ async def escalation_wrapper(
...
+            for k, v in serialized_input.items()
+            if v is not None
+        ]
+        if not fields:
+            return None
+
+        request = MemorySearchRequest(
+            fields=fields,
+            settings=SearchSettings(
+                threshold=0.0, result_count=1, search_mode=SearchMode.Hybrid
```
swathiJayav: SearchSettings should be read from user's Memory settings. Except for `result_count` - this is always 1 for escalation memory.

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -333,3 +363,98 @@ async def escalation_wrapper(
...
+        )
+        if response.results and response.results[0].answer:
+            cached = response.results[0].answer
+            _escalation_logger.info(
+                "Escalation memory cache hit for space '%s'", memory_space_id
+            )
+            return {
+                "action": EscalationAction.CONTINUE,
+                "output": cached.output,
+                "outcome": cached.outcome or "cached",
```
swathiJayav: This outcome is tied to the Escalation and configured by user to be something like Submit, Reject, etc. We can't add our own default here.

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -333,3 +363,98 @@ async def escalation_wrapper(
...
+
+    return None
+
+
+async def _ingest_escalation_memory(
+    memory_space_id: str | None,
+    answer: str,
+    attributes: str,
+    span_id: str = "",
+    trace_id: str = "",
```
swathiJayav: Shouldn't be "nullable"

src/uipath_langchain/agent/react/types.py
```diff
@@ -55,6 +56,23 @@ class AgentGraphNode(StrEnum):
...
+    INIT that queries the memory service and stores the server-generated
+    systemPromptInjection in ``inner_state.memory_injection``.
+    """
+
+    memory_space_id: str = Field(description="GUID of the memory space to query.")
+    folder_key: str | None = Field(
+        default=None, description="Folder key for the memory resource."
+    )
+    result_count: int = Field(default=5, ge=1, le=10)
+    threshold: float = Field(default=0.0, ge=0.0, le=1.0)
```
swathiJayav, line 83: Pick something beteween defaults set in FE for both Episodic/Escalation memory.

src/uipath_langchain/agent/react/memory_node.py
```diff
@@ -0,0 +1,113 @@
...
+            settings=SearchSettings(
+                threshold=memory_config.threshold,
+                result_count=memory_config.result_count,
+                search_mode=SearchMode.Hybrid,
+            ),
+        )
+
+        try:
+            sdk = UiPath()
+            response = await sdk.memory.search_async(
```
swathiJayav: Search and Ingest for both episodic and escalation memory need to be scoped on FolderKey, similar to Temporal-based Agents

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -262,6 +285,13 @@ async def create_escalation_task():
             EscalationAction(outcome_str) if outcome_str else EscalationAction.CONTINUE
         )
 
+        # --- Escalation memory: persist outcome for future recall ---
+        await _ingest_escalation_memory(
```
swathiJayav: pass spanId, traceId, userId here

src/uipath_langchain/agent/react/types.py
```diff
@@ -55,6 +56,23 @@ class AgentGraphNode(StrEnum):
...
+
+class MemoryConfig(BaseModel):
+    """Configuration for Agent Episodic Memory.
+
+    When passed to ``create_agent()``, a MEMORY_RECALL node is added before
+    INIT that queries the memory service and stores the server-generated
+    systemPromptInjection in ``inner_state.memory_injection``.
+    """
+
+    memory_space_id: str = Field(description="GUID of the memory space to query.")
```
swathiJayav, line 70: What about weights for each of the fields added to memory? For episodic memory, these are agent input fields. For escalation memory, these are the input fields set for the escalation. Not all fields are used for memory search, the user picks which ones and sets a weight for each.

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -333,3 +363,98 @@ async def escalation_wrapper(
...
+            ),
+        )
+        sdk = UiPath()
+        response = await sdk.memory.escalation_search_async(
+            memory_space_id=memory_space_id, request=request
+        )
+        if response.results and response.results[0].answer:
+            cached = response.results[0].answer
+            _escalation_logger.info(
+                "Escalation memory cache hit for space '%s'", memory_space_id
```
swathiJayav: Should also add Span for found from Memory

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -262,6 +285,13 @@ async def create_escalation_task():
...
         )
 
+        # --- Escalation memory: persist outcome for future recall ---
+        await _ingest_escalation_memory(
+            _memory_space_id,
+            answer=json.dumps(escalation_output),
+            attributes=json.dumps(serialized_data),
+        )
+
         return {
```
swathiJayav, line 265: Add saveToMemory=true in span

src/uipath_langchain/agent/react/types.py
mjnovice: Addressed — added `field_weights: dict[str, float]` to `MemoryConfig`. When specified, only listed fields are included in the search (matching backend behavior), each with its configured weight. Empty dict = all fields with default weight for backward compatibility. See 6ca8307.

src/uipath_langchain/agent/react/memory_node.py
mjnovice: Fixed — both `_check_escalation_memory_cache` and `_ingest_escalation_memory` now accept `folder_path` from the execution context (`get_execution_folder_path()`), resolve it to a `folder_key` via `sdk.folders.retrieve_folder_key()`, and pass it explicitly to the SDK's `escalation_search_async` / `escalation_ingest_async` calls. This matches the Temporal backend which scopes all memory operations on FolderKey. See e355e19.

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Fixed — the answer stored in escalation memory now includes both output and outcome as `{"output": ..., "outcome": ...}`, matching the Temporal backend's serialization (`new { taskResult.Output, taskResult.Outcome }`). Attributes are also wrapped in `{"arguments": ...}` to match the backend's `new JsonObject { ["arguments"] = payload.Input.Arguments }`. See 69e6e1c.

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Already addressed in 69e6e1c — attributes are now wrapped as `{"arguments": serialized_data}`, matching the Temporal backend's `new JsonObject { ["arguments"] = payload.Input.Arguments }` at line 503.

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Fixed in b357529 — SearchSettings (threshold, searchMode) are now read from the escalation resource's `properties.memory` config instead of being hardcoded. Field-level weights from `fieldSettings` are also applied, and only configured fields are included in the search (matching `BuildMemorySearchRequest` in `EscalationToolExecutor.cs:714-747`). `result_count` stays fixed at 1.

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Fixed in bc7d092 — now passing `spanId` and `traceId` from the current OpenTelemetry span context (with fallback to `UIPATH_TRACE_ID` env var). `userId` is left as None for now since the Python escalation result doesn't carry `ReviewedBy` — the Temporal backend resolves it via identity client (`EmailResolveAsync`, line 514-516), which isn't available here yet.

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Added in 98216c4 — `savedToMemory=true` is set on the OTel span after successful ingest, `false` on failure. Also sets `fromMemory=false` when going through the ingest path. Refs `EscalationToolSpanAttributes.cs:19-25` and `EscalationToolWorkflow.cs:131-133`.

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Added in 98216c4 — `fromMemory=true` is set on the OTel span on cache hit in `_check_escalation_memory_cache`. Refs `EscalationToolWorkflow.cs:103`.

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Fixed in 72c07f9 — removed the `or "cached"` fallback. The outcome is now passed through as-is from the cache, matching the Temporal backend which uses `cachedRecall.Outcome` without any default (EscalationToolExecutor.cs:670).

src/uipath_langchain/agent/react/types.py
mjnovice: Fixed in 849c573 — changed `result_count` default from 5 to 3 to match the FE episodic memory default (`agentEditor.ts:324-328`). `threshold=0.0` already matched. FE defaults: episodic threshold=0, k=3; escalation threshold=0.8, k=1 (hardcoded).

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Fixed in 3b54ae5 — `memory_space_id` is now `str` (non-nullable) in both `_check_escalation_memory_cache` and `_ingest_escalation_memory`. The caller guards with `if _memory_space_id:` before calling.

src/uipath_langchain/agent/tools/escalation_tool.py
swathiJayav: Is there a timeline for when Identity support will be added? We want to show the ReviwedBy in our Instance Management page in the near future. We need to store the user info with the memories so that all memories are associated with a user when we do this.

src/uipath_langchain/agent/react/types.py
swathiJayav: We need to validate that at least one field + weight is added here. It is not optional.

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -333,3 +396,194 @@ async def escalation_wrapper(
     tool.set_tool_wrappers(awrapper=escalation_wrapper)
 
     return tool
+
+
+# --- Escalation memory helpers ---
+
+
+def _get_current_span_and_trace_ids() -> tuple[str, str]:
```
swathiJayav: This should be pulled out to a common utils or library. Does something like this already exist since other tools would need to do the same formatting?

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Update: userId is now populated in 56a41d9. We extract `completed_by_user` from the raw escalation result (before `EscalationToolOutput` validation drops extra fields), then use the existing `_get_user_email()` helper to get the email address. This is passed as `user_id` to the ingest request. Note: the Temporal backend resolves email → user ID via `EmailResolveAsync`, but we pass the email directly since the SDK's `user_id` field accepts it.

mjnovice: My bad, claude cheated here. 

It seems there was a way to add userId, I added it.

src/uipath_langchain/agent/react/types.py
mjnovice: Fixed in 778bb52 — `field_weights` is now required (no default) with a model validator ensuring at least one entry. Empty dict raises `ValueError: field_weights must contain at least one field`.

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: Addressed in 3032bec — moved both `get_current_span_and_trace_ids()` and `set_span_attribute()` to `_utils/_otel.py` and exported them from `_utils/__init__.py`. Other tools can now import from `uipath_langchain._utils` instead of reimplementing.

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -333,3 +407,162 @@ async def escalation_wrapper(
...
+            memory_space_id=memory_space_id,
+            request=request,
+            folder_key=folder_key,
+        )
+        if response.results and response.results[0].answer:
+            cached = response.results[0].answer
+            _escalation_logger.info(
+                "Escalation memory cache hit for space '%s'", memory_space_id
+            )
+            # Ref: EscalationToolWorkflow.cs:103 — span.Attributes.FromMemory = true
```
andreitava-uip: no need to leave comments referencing private repos in a public library. 
Applicable in all other comments.

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -333,3 +407,162 @@ async def escalation_wrapper(
     tool.set_tool_wrappers(awrapper=escalation_wrapper)
 
     return tool
+
+
+# --- Escalation memory helpers ---
+
+
+async def _check_escalation_memory_cache(
```
andreitava-uip: i'd move the memory helpers to a separate file since they are completely standalone

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -333,3 +407,162 @@ async def escalation_wrapper(
...
+        )
+        if response.results and response.results[0].answer:
+            cached = response.results[0].answer
+            _escalation_logger.info(
+                "Escalation memory cache hit for space '%s'", memory_space_id
+            )
+            # Ref: EscalationToolWorkflow.cs:103 — span.Attributes.FromMemory = true
+            set_span_attribute("fromMemory", True)
+            return {
+                "action": EscalationAction.CONTINUE,
```
andreitava-uip: I don't think hardcoding `CONTINUE` is correct. Do we not commit to memory other outcomes?

src/uipath_langchain/agent/react/memory_node.py
```diff
@@ -0,0 +1,177 @@
...
+    On failure it logs a warning and continues with an empty injection.
+
+    Args:
+        memory_config: Memory configuration with space ID and search settings.
+
+    Returns:
+        An async callable suitable for ``builder.add_node()``.
+    """
+
+    async def memory_recall_node(state: Any) -> dict[str, Any]:
```
andreitava-uip, line 52: have we tried to type the state?

src/uipath_langchain/agent/react/init_node.py
```diff
@@ -24,6 +24,19 @@ def graph_state_init(state: Any) -> Any:
...
+        if hasattr(state, "inner_state") and hasattr(
+            state.inner_state, "memory_injection"
+        ):
+            memory_injection = state.inner_state.memory_injection or ""
+        if memory_injection and resolved_messages:
+            first = resolved_messages[0]
+            if isinstance(first, SystemMessage):
+                resolved_messages[0] = SystemMessage(
+                    content=str(first.content) + memory_injection
+                )
```
andreitava-uip, line 40: I'd say we should clearly delimit the memory section in the prompt. Currently it just gets appended.

src/uipath_langchain/_utils/_otel.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def get_current_span_and_trace_ids() -> tuple[str, str]:
+    """Get current OpenTelemetry span ID and trace ID.
+
+    Returns hex-encoded IDs, or empty strings if no active span.
+    Falls back to UIPATH_TRACE_ID env var for trace_id.
+    """
+    try:
+        from opentelemetry import trace
```
andreitava-uip: any reason to lazy-load this? It's loaded in every utility form this file.

src/uipath_langchain/agent/react/memory_node.py
```diff
@@ -0,0 +1,177 @@
...
+            return {}
+
+        fields = _build_search_fields(
+            input_arguments, field_weights=memory_config.field_weights or None
+        )
+        if not fields:
+            logger.warning(
+                "Memory recall: no search fields after filtering (inputs=%s, weights=%s)",
+                list(input_arguments.keys()),
+                memory_config.field_weights,
```
andreitava-uip: I think you forgot to lower these logs. In my opinion they should not be warnings because memory misses are not warning worthy but expected in normal execution.

src/uipath_langchain/agent/react/memory_node.py
```diff
@@ -0,0 +1,177 @@
...
+            if not folder_key and memory_config.folder_path:
+                folder_key = sdk.folders.retrieve_folder_key(
+                    memory_config.folder_path
+                )
+            logger.warning(
+                "Memory recall: searching space='%s', folder_key='%s', "
+                "fields=%s, threshold=%s, result_count=%s",
+                memory_config.memory_space_id,
+                folder_key,
+                [(f.key_path, f.value) for f in fields],
```
andreitava-uip: do we have no concerns loggin the raw value? Could include PII. Are we ok with PII in robot logs?

src/uipath_langchain/agent/react/memory_node.py
```diff
@@ -0,0 +1,177 @@
...
+                        error_detail = f"{exc} | status={resp.status_code} body={resp.text}"
+                    except Exception:
+                        pass
+                    break
+            logger.warning(
+                "Memory recall failed for space '%s': %s",
+                memory_config.memory_space_id,
+                error_detail,
+            )
+            injection = ""
```
andreitava-uip: doesn't the sdk client raise an EnrichedException here? It should already have typed info on it.

src/uipath_langchain/agent/react/memory_node.py
```diff
@@ -0,0 +1,177 @@
...
+
+        if not injection:
+            return {}
+
+        return {"inner_state": {"memory_injection": injection}}
+
+    return memory_recall_node
+
+
+def _extract_user_inputs(state: Any) -> dict[str, Any]:
```
andreitava-uip: we already have `extract_input_data_from_state` in `utils.py`

src/uipath_langchain/agent/tools/escalation_tool.py
```diff
@@ -333,3 +407,162 @@ async def escalation_wrapper(
...
+    """
+
+    try:
+        from uipath.platform.memory import (
+            FieldSettings,
+            MemorySearchRequest,
+            SearchField,
+            SearchMode,
+            SearchSettings,
+        )
```
andreitava-uip: any reason for the lazy load?

tests/agent/tools/test_escalation_memory.py
```diff
@@ -0,0 +1,123 @@
...
+        mock_response = MagicMock()
+        mock_response.results = [mock_match]
+        mock_sdk.memory.escalation_search_async = AsyncMock(return_value=mock_response)
+
+        result = await _check_escalation_memory_cache(
+            "space-123", {"Content": "Is the sky blue?"}
+        )
+
+        assert result is not None
+        assert result["action"] == EscalationAction.CONTINUE
```
andreitava-uip: this mirrors my other comment over whether the action result should get hardcoded to CONTINUE

andreitava-uip (changes requested):

src/uipath_langchain/_utils/_otel.py
```diff
@@ -0,0 +1,39 @@
+"""OpenTelemetry utilities for span and trace context."""
+
+import os
+from typing import Any
+
+
+def get_current_span_and_trace_ids() -> tuple[str, str]:
```
AAgnihotry: why do you need this, we already has a function to get current span

src/uipath_langchain/agent/react/_memory_tracing.py
```diff
@@ -0,0 +1,162 @@
...
+        "Source": 4,  # Python SDK source
+        "OrganizationId": os.environ.get("UIPATH_ORGANIZATION_ID"),
+        "TenantId": os.environ.get("UIPATH_TENANT_ID"),
+        "FolderKey": os.environ.get("UIPATH_FOLDER_KEY"),
+        "ProcessKey": os.environ.get("UIPATH_PROCESS_UUID"),
+        "JobKey": os.environ.get("UIPATH_JOB_KEY"),
+    }
+
+
+async def emit_memory_recall_spans(
```
AAgnihotry: why all of these in langchain python ?

src/uipath_langchain/agent/tools/escalation_tool.py
mjnovice: yeah, this was more for my own reference as well as claude. Might still want to keep it, if not a biggie. Will remove it before merging

src/uipath_langchain/agent/react/init_node.py
mjnovice: This is in parity with the C# backend. In `BasicAgent.cs:47-58`, the `systemPromptInjection` is substituted raw via `template.Replace("{{fewShotExamples}}", ...)` — no additional delimiters or section markers are added by the backend. Any formatting/delimiters are the responsibility of the LLmOps API response itself.

Our approach appends the injection string as-is to the system message, matching this behavior.

src/uipath_langchain/agent/react/memory_node.py
mjnovice: Fixed — the verbose log line that included raw field values has been removed. Remaining logs only contain metadata (space ID, result count, field names) with no raw user input values.

src/uipath_langchain/agent/react/memory_node.py
mjnovice: Fixed — replaced `_extract_user_inputs` with the existing `extract_input_data_from_state` from `utils.py`. The `input_schema` is now passed through from `create_agent()` → `create_memory_recall_node()`.

src/uipath_langchain/_utils/_otel.py
mjnovice: The lazy import is intentional — `opentelemetry` is an optional dependency, so the `try/except ImportError` pattern handles environments where it's not installed. Moving it to the top level would cause an import error when OTel isn't available.

src/uipath_langchain/agent/react/memory_node.py
mjnovice: I tried, it was getting tedious. I will revisit this later.

mjnovice: https://github.com/UiPath/uipath-langchain-python/issues/806

src/uipath_langchain/agent/react/_memory_tracing.py
mjnovice: Removed — `_memory_tracing.py` was dead code. It was an earlier approach to manual span emission that got superseded by the OTel `tracer.start_as_current_span` calls in `memory_node.py`. File deleted.

src/uipath_langchain/agent/react/memory_node.py
```diff
@@ -0,0 +1,224 @@
...
+            )
+
+            with fewshot_span_ctx as fewshot_span:
+                try:
+                    sdk = UiPath()
+                    folder_key = memory_config.folder_key
+                    if not folder_key and memory_config.folder_path:
+                        folder_key = sdk.folders.retrieve_folder_key(
+                            memory_config.folder_path
+                        )
```
andreitava-uip: nit: I think there is an async variant for `retrieve_folder_key`. We should probably use that so we don't block the event loop.

src/uipath_langchain/agent/react/memory_node.py
```diff
@@ -0,0 +1,224 @@
...
+                        error_detail = repr(e)
+                    logger.warning(
+                        "Memory recall failed for space '%s': %s",
+                        memory_config.memory_space_id,
+                        error_detail,
+                    )
+                    if lookup_span and hasattr(lookup_span, "set_status"):
+                        lookup_span.set_status(
+                            otel_trace.StatusCode.ERROR, error_detail
+                        )
```
andreitava-uip: why import EnrichedException locally? The module is not heavy/optional.
The only thing this protects against is if there's some import issues with uipath_platform, in which case the original error is masked as the handler itself will raise `ImportError`.

src/uipath_langchain/agent/react/memory_node.py
```diff
@@ -0,0 +1,224 @@
...
+
+            # Set result attributes after search completes
+            if lookup_span and hasattr(lookup_span, "set_attribute"):
+                lookup_span.set_attribute("memoryItemsMatched", results_count)
+                if injection:
+                    lookup_span.set_attribute("result", injection)
```
andreitava-uip, line 193: shouldn't the outer fewshot_span status be marked on error as well?

tests/agent/react/test_memory_node.py
```diff
@@ -0,0 +1,186 @@
...
+        class TestInput(BM):
+            topic: str = Field(default="")
+            level: str = Field(default="")
+
+        state = {
+            "messages": [],
+            "inner_state": {},
+            "topic": "python",
+            "level": "advanced",
+        }
```
andreitava-uip: nit: could be worth to have a single utility for quickly creating the state object, ideally referencing the actual pydantic model we use so drifts are caught early.

andreitava-uip (changes requested):

tests/agent/react/test_memory_node.py
```diff
@@ -0,0 +1,186 @@
...
+        )
+        assert len(fields) == 1
+        assert fields[0].key_path == ["agent-input", "topic"]
+
+    def test_empty_input(self) -> None:
+        assert _build_search_fields({}) == []
+
+
+class TestExtractUserInputs:
+    def test_filters_internal_fields(self) -> None:
```
andreitava-uip: nit: this belongs in test_utils.py not here

tests/agent/react/test_memory_node.py
```diff
@@ -0,0 +1,186 @@
...
+
+def _make_mock_model() -> MagicMock:
+    return MagicMock(spec=BaseChatModel)
+
+
+def _make_mock_tool(name: str = "test_tool") -> Mock:
+    tool = Mock(spec=BaseTool)
+    tool.name = name
+    return tool
+
```
andreitava-uip: is this used?

tests/agent/react/test_memory_node.py
```diff
@@ -0,0 +1,186 @@
...
+        assert fields[0].key_path == ["agent-input", "topic"]
+
+    def test_empty_input(self) -> None:
+        assert _build_search_fields({}) == []
+
+
+class TestExtractUserInputs:
+    def test_filters_internal_fields(self) -> None:
+        from pydantic import BaseModel as BM
+        from pydantic import Field
```
andreitava-uip: No need to locally import. In fact we already import `BaseModel `at module level.

src/uipath_langchain/agent/react/memory_node.py
mjnovice: Fixed — switched to `await sdk.folders.retrieve_folder_key_async()` to avoid blocking the event loop. See 8c8b6b1d.

mjnovice: Fixed — switched to `await sdk.folders.retrieve_folder_key_async()`.

src/uipath_langchain/agent/react/memory_node.py
mjnovice: Fixed — moved `EnrichedException` to top-level import.

src/uipath_langchain/agent/react/memory_node.py
mjnovice: Fixed — now setting error status on both `fewshot_span` and `lookup_span`.

tests/agent/react/test_memory_node.py
mjnovice: Fixed — added `_make_state()` helper and reusable `_TopicInput`/`_TopicLevelInput` models.

tests/agent/react/test_memory_node.py
mjnovice: Fixed — removed `TestExtractUserInputs` (belongs in test_utils.py) and the local import.

tests/agent/react/test_memory_node.py
mjnovice: Removed — `_make_mock_tool` was unused.

tests/agent/react/test_memory_node.py
mjnovice: Fixed — removed local import, using module-level `BaseModel`.

andreitava-uip (approved):

PR #794, opened 2026-04-22
Feat/file masking

yashwagle1:


pyproject.toml
saksharthakkar: you'll need to bump up this too

src/uipath_langchain/agent/tools/internal_tools/analyze_files_tool.py
```diff
@@ -81,6 +205,23 @@ async def tool_fn(**kwargs: Any):
...
+            logger.exception("Failed to fetch deployed policy")
+
+        pii_result: PiiDetectionResponse | None = None
+        if client is not None and is_pii_policy_enabled(policy):
+            try:
+                analysis_task, files, pii_result = await _apply_pii_masking(
+                    client, policy, analysis_task, files
+                )
+            except Exception as e:
+                logger.error("PII detection raised: %r", e, exc_info=True)
```
saksharthakkar: this will catch exception, log and continue? Is that expected? or should it stop on exception?

saksharthakkar:
please add tests

src/uipath_langchain/agent/tools/internal_tools/analyze_files_tool.py
```diff
@@ -38,12 +48,126 @@
...
+       (defaults to ``True``; override via ``FeatureFlags.configure_flags``
+       or ``UIPATH_FEATURE_File-Pii-Masking-Enabled`` env var).
+    2. Platform policy — ``data.container.pii-in-flight-agents`` from the
+       AutomationOps deployed-policy response.
+    """
+    if not FeatureFlags.is_flag_enabled(_PII_MASKING_FEATURE_FLAG, default=True):
+        return False
+    if not policy:
+        return False
+    container = policy.get("data", {}).get("container", {})
```
saksharthakkar: can this be cached? instead of doing get policy on each call?

yashwagle1: That is the plan to cache it in agenthub side. Should we cache it in the agent as well?

cristian-groza (approved):

src/uipath_langchain/agent/tools/internal_tools/analyze_files_tool.py
saksharthakkar: that works!

src/uipath_langchain/agent/tools/internal_tools/analyze_files_tool.py
```diff
@@ -102,6 +127,13 @@ async def tool_fn(**kwargs: Any):
         del messages, human_message_with_files, files
 
         analysis_result = extract_text_content(result)
+
+        if masker is not None:
+            try:
+                analysis_result = masker.rehydrate(analysis_result)
+            except Exception:
+                logger.exception("Failed to rehydrate LLM response")
```
saksharthakkar: if rehydrate fails, we should throw exception?

saksharthakkar (approved):

src/uipath_langchain/agent/tools/internal_tools/analyze_files_tool.py
yashwagle1: Threw an exception

PR #810, opened 2026-04-24
fix: make MemoryConfig.field_weights optional

mjnovice:
## Summary

- Make `field_weights` optional (`dict[str, float] | None`, default `None`) instead of required with a strict validator
- Remove the `_validate_field_weights` model validator that raised `ValueError` for empty dicts

### Why

When no `fieldSettings` are configured in the agent definition's `dynamicFewShotSettings`, the runtime passes `field_weights={}` which fails validation. In practice, memory spaces may not have field weights configured — in that case, all input fields should be searched with default weights.

This was causing `AGENT_STARTUP.UNEXPECTED_ERROR` in production:
```
1 validation error for MemoryConfig
  Value error, field_weights must contain at least one field
```

## Test plan

- [x] Agents with no fieldSettings configured can start without error
- [x] Agents with fieldSettings still work as before (weights are passed through)


mjnovice:
closing this as this is not required anymore

PR #814, opened 2026-04-28
fix: honor checkpointer.serde.allowed_msgpack_modules in langgraph.json (#1500)

### LINKED ISSUE #1500: [DX] GAP-07: SDK's own CreateTask type not registered with msgpack, produces warning on every resume
## Summary

Every `--resume` run prints a noisy deserialization warning for `CreateTask`, which is a first-party SDK type. The warning cannot be silenced without patching the SDK or suppressing log output globally.

## Observed Output

```
Deserializing unregistered type uipath.platform.common.interrupt_models.CreateTask
from checkpoint. This will be blocked in a future version.
Add to allowed_msgpack_modules to silence: [('uipath.platform.common.interrupt_models', 'CreateTask')]
```

## Root Cause

`CreateTask` is defined in `uipath.platform.common.interrupt_models` but is not auto-registered with the msgpack serializer used for checkpoint persistence. LangGraph's msgpack serializer requires types to be explicitly registered in `allowed_msgpack_modules` to deserialize them without a warning.

Since `CreateTask` is a first-party SDK type that the SDK itself places in checkpoints (via `interrupt()`), users should never have to manually register it.

## Workaround

None without patching the SDK or suppressing the warning at the logging level. The suggested fix in the warning message (`allowed_msgpack_modules`) would require the user to configure LangGraph internals.

## Suggested Fix

Auto-register all first-party interrupt model types (at minimum `CreateTask`, and any other types from `uipath.platform.common.interrupt_models`) with msgpack on import. This is a one-line fix:

```python
# In the SDK's initialization code
from langgraph.checkpoint.serde.jsonplus import JsonPlusSerializer
# or equivalent msgpack registration
```

This warning appears on every resume and buries actual errors in the log output, making it harder to spot real problems.

## Impact

- Severity: **Low** (functional but noisy)
- Every user of HITL with `--resume` sees this warning
- The warning says "This will be blocked in a future version" — when that happens, this becomes a breaking change for all HITL agents

radu-mocanu:
## Summary
- read `checkpointer.serde.allowed_msgpack_modules` from `langgraph.json` and apply it to the saver's serializer
- union the user's list with the SDK's first-party interrupt models so checkpoint payloads like `CreateTask` and `InvokeProcess` keep round-tripping silently
- behavior is opt-in: with no `serde` block, langgraph's default permissive serde is preserved
- adds the `samples/checkpoint-allowlist` sample and the `testcases/checkpoint-allowlist` integration test that asserts no `Deserializing unregistered type` warning fires when the block is set

## Why

Fixes UiPath/uipath-python#1500

users couldn't silence the `Deserializing unregistered type ...CreateTask` warning that fires on every HITL `--resume` because uipath-langchain owns saver construction and didn't expose any hook. this PR matches the public schema documented for the langgraph CLI so users can declare their custom types alongside ours.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-langchain"
-version = "0.10.10"
+version = "0.10.11"
 description = "Python SDK that enables developers to build and deploy LangGraph agents to the UiPath Cloud Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
--- a/samples/checkpoint-allowlist/README.md
+++ b/samples/checkpoint-allowlist/README.md
@@ -1,0 +1,15 @@
+# checkpoint-allowlist
+
+Shows how to silence langgraph's `Deserializing unregistered type` warning by declaring custom types in `langgraph.json`'s `serde.allowed_msgpack_modules` block.
+
+Graph: `START -> evaluate -> finalize -> END`. State carries a custom Pydantic `Score` value, which langgraph reconstructs on every checkpoint round-trip.
+
+## Run
+
+```bash
+uv sync
+uv run uipath init
+uv run uipath run agent --file input.json
+```
+
+Expect: agent completes, no `Deserializing unregistered type ...Score` warning. Remove the `serde` block from `langgraph.json` to see the warning return.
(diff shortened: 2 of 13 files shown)


radugheo (approved):

radu-mocanu (commented):
Langgraph feature request: https://github.com/langchain-ai/langgraph/issues/7695

PR #819, opened 2026-05-01
feat: add client side tools to mapper and runtime

norman-le:
```
## uipath-python

- Added ClientSide tool type to agent model so the runtime can distinguish client-side tools from server tools
- Bridge emits executingToolCall event on durable interrupt when is_execution_phase marker is set, enabling the client to know when to execute the tool
- Bridge routes endToolCall from the client back to the agent, completing the interrupt/resume cycle

## uipath-langchain-python

- Client-side tool factory creates StructuredTool instances that use @durable_interrupt to pause the graph and wait for client execution
- @mockable wraps @durable_interrupt (as a separate outer function) so evals can simulate client-side tools without hitting GraphInterrupt
- CLIENT_SIDE_TOOL_MARKER defined in hitl.py as the single source of truth for identifying client-side tools across the codebase
- Error-handling wrapper (RunnableCallableWithTool) now preserves .tool reference for client-side tools, enabling runtime discovery
- Runtime discovers client-side tools from the compiled graph via shared _iter_graph_tools() iterator, used by both confirmation and client-side tool detection
- MessageMapper emits executingToolCall only for server tools without interrupts; client-side and confirmation tools get it from the bridge instead
- MessageMapper suppresses endToolCall for client-side tool results since the client already produced the result
- MessageMapper sets isClientSideTool and outputSchema on startToolCall events so the client knows to show the input form
- 9 new tests covering tool discovery, executingToolCall emission rules, and endToolCall suppression

## uipath-agents-python

- Graph builder strips CLIENT_SIDE_TOOL_MARKER during evals so @durable_interrupt is bypassed and @mockable can intercept

## AgentInterfaces

- CAS tracks is_client_side_tool on tool calls in the database and forwards client-originated endToolCall to the agent
- CAS validates client-side tool declarations against the agent definition and guards against unauthorized endToolCall for non-client-side tools
- React SDK listens for executingToolCall events and renders a form (ClientSideToolWidget) using the tool's output schema for the user to fill in
- ClientSideToolWidget reuses the existing ToolConfirmationFormContent component
- Added type schemas for executingToolCall events and client-side tool declarations
```

Videos in CAS localhost (Agent Builder changes already published, but this will be the debug experience as well when the sdk is upgraded):

Without tool confirmation for client side tool:

https://github.com/user-attachments/assets/48bc0979-0f51-471c-90aa-af561d876c31

With tool confirmation for client side tool:

https://github.com/user-attachments/assets/384ba36d-db2d-478e-8a74-8b3de0501a06

URT Eval:
image

Related to changes in other PRs:
CAS: https://github.com/UiPath/AgentInterfaces/pull/949
uipath-langchain-python (tool definition): https://github.com/UiPath/uipath-langchain-python/pull/819
uipath-python (bridge changes): https://github.com/UiPath/uipath-python/pull/1609
uipath-agents-python (skipping for evals, so it can be mocked): https://github.com/UiPath/uipath-agents-python/pull/485
Another Agents PR fixing some issues with debug: https://github.com/UiPath/Agents/pull/5250

### DIFF
--- a/src/uipath_langchain/agent/tools/client_side_tool.py
+++ b/src/uipath_langchain/agent/tools/client_side_tool.py
@@ -1,0 +1,95 @@
+"""Factory for creating client-side tools that execute on the client SDK."""
+
+import inspect
+import json
+from logging import getLogger
+from typing import Annotated, Any
+
+from langchain_core.messages import ToolMessage
+from langchain_core.tools import InjectedToolCallId, StructuredTool
+from uipath.agent.models.agent import AgentClientSideToolResourceConfig
+from uipath.eval.mocks import mockable
+
+from uipath_langchain._utils.durable_interrupt import durable_interrupt
+from uipath_langchain.agent.react.jsonschema_pydantic_converter import (
+    create_model as create_model_from_schema,
+)
+
+from .utils import sanitize_tool_name
+
+logger = getLogger(__name__)
+
+CLIENT_SIDE_TOOL_MARKER = "uipath_client_tool"
+
+
+def create_client_side_tool(
+    resource: AgentClientSideToolResourceConfig,
+) -> StructuredTool:
+    """Create a client-side tool that pauses the graph and waits for the client to execute it.
+
+    The tool uses @durable_interrupt to suspend the graph. The client SDK receives
+    an executingToolCall event, runs its registered handler, and sends endToolCall
+    back through CAS. The bridge routes that endToolCall to wait_for_resume(),
+    which unblocks the graph with the client's result.
+    """
+    tool_name = sanitize_tool_name(resource.name)
+    input_model = create_model_from_schema(resource.input_schema)
+
+    async def client_side_tool_fn(
+        *, tool_call_id: Annotated[str, InjectedToolCallId], **kwargs: Any
+    ) -> Any:
+        @mockable(
+            name=resource.name,
+            description=resource.description,
+            input_schema=input_model.model_json_schema(),
+            output_schema=(resource.output_schema or {}),
+            example_calls=getattr(resource.properties, 'example_calls', None),
+        )
+        @durable_interrupt
+        async def wait_for_client_execution() -> dict[str, Any]:
+            return {
+                "tool_call_id": tool_call_id,
+                "tool_name": tool_name,
+                "input": kwargs,
+                "is_execution_phase": True,
+            }
+
+        # First run: raises GraphInterrupt with the tool call info.
+        # On resume: returns the client's result (output, isError, etc.)
+        # During evals: @mockable intercepts and returns simulated response.
+        result = await wait_for_client_execution()
+
+        # The resume value from the bridge is the endToolCall payload
+        output = result.get("output")
+        is_error = result.get("is_error", False)
+
+        content = str(output) if output is not None else ""
+        if isinstance(output, dict):
+            content = json.dumps(output)
+
+        return ToolMessage(
+            content=content,
+            tool_call_id=tool_call_id,
+            status="error" if is_error else "success",
+            response_metadata={CLIENT_SIDE_TOOL_MARKER: True},
+        )
+
+    # Patch signature so LangChain injects tool_call_id at runtime
+    original_sig = inspect.signature(client_side_tool_fn)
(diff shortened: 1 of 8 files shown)


src/uipath_langchain/agent/tools/client_side_tool.py
```diff
@@ -0,0 +1,95 @@
...
+            status="error" if is_error else "success",
+            response_metadata={CLIENT_SIDE_TOOL_MARKER: True},
+        )
+
+    # Patch signature so LangChain injects tool_call_id at runtime
+    original_sig = inspect.signature(client_side_tool_fn)
+    params = [p for p in original_sig.parameters.values() if p.name != "kwargs"] + [
+        inspect.Parameter("kwargs", inspect.Parameter.VAR_KEYWORD, annotation=Any),
+    ]
+    client_side_tool_fn.__signature__ = original_sig.replace(parameters=params)
```
JoshParkSJ: Hm a bit confused what this is for

src/uipath_langchain/runtime/messages.py
```diff
@@ -476,19 +502,24 @@ async def map_tool_message_to_events(
...
-            UiPathConversationMessageEvent(
-                message_id=message_id,
-                tool_call=UiPathConversationToolCallEvent(
-                    tool_call_id=message.tool_call_id,
-                    end=UiPathConversationToolCallEndEvent(
-                        timestamp=self.get_timestamp(),
-                        output=content_value,
-                        is_error=message.status == "error",
+        # Suppress endToolCall for client-side tools — the client already has the result (it produced it).
+        is_client_side = message.response_metadata.get("uipath_client_tool", False)
```
JoshParkSJ: use `CLIENT_SIDE_TOOL_MARKER`

src/uipath_langchain/agent/tools/client_side_tool.py
```diff
@@ -0,0 +1,85 @@
...
+            response_metadata={CLIENT_SIDE_TOOL_MARKER: True},
+        )
+
+    tool = StructuredTool(
+        name=tool_name,
+        description=resource.description or f"Client-side tool: {tool_name}",
+        args_schema=input_model,
+        coroutine=client_side_tool_fn,
+        metadata={
+            CLIENT_SIDE_TOOL_MARKER: True,
```
maxduu, line 80: Maybe a more consistent name compared to the other markers, which include `CONVERSATIONAL`? I'm thinking `IS_CONVERSATIONAL_CLIENT_SIDE_TOOL`?

src/uipath_langchain/runtime/messages.py
```diff
@@ -436,15 +440,40 @@ async def map_current_message_to_start_tool_call_events(self):
...
+                        # get executingToolCall from the bridge instead.
+                        if not require_confirmation and not is_client_side:
+                            events.append(
+                                UiPathConversationMessageEvent(
+                                    message_id=self.current_message.id,
+                                    tool_call=UiPathConversationToolCallEvent(
+                                        tool_call_id=tool_call["id"],
+                                        executing=UiPathConversationExecutingToolCallEvent(
+                                            tool_name=tool_call["name"],
+                                            input=tool_call["args"],
```
maxduu, line 471: Hm, I guess related to my comment [here](https://github.com/UiPath/uipath-python/pull/1609/changes#r3243912306). Personally would prefer if a single place creates/outputs the executingToolCall event, especially since right now it seems to be split across two repos and could add some confusion. Is there a code-place where any path (including tools with interrupts) eventually converge to once they are ready to run the tool?

norman-le: I wasnt able to get them to fire in a single place, because the MessageMapper processes the AIMessage before the tool node runs (MessageMapper sees tool calls at stream time), but the graph isn't ready to receive responses until interrupt time. For tools that don't need a response from the user, this doesn't matter. For tools that pause and wait, the emission must happen after the pause (tool confirmation and client side tools).

I added it in both places to keep the events consistent for tool calls, but it's a no-op for non interrupt tools anyways. I think if it's causing issues, removing it from non interrupt tools might be the better case so that users wont get confused and try to handle that event.

PR #821, opened 2026-05-04
fix: ixp extraction tool

cristian-groza:
Summary                                                                                                                                                                                                                           
                                                                                                                                                                                                                                    
  Fixes broken IXP extraction tool invocation caused by LangChain's mishandling of pydantic field aliases.                                                                                                                          
   
  Problem                                                                                                                                                                                                                           
                  
  After PR #796 switched the extraction tool's args_schema to Attachment, tool calls started failing — kwargs.get("id") and kwargs.get("full_name") came back as None, so attachments.download_async was called with key=None.

  Two LangChain bugs are at play:

  1. BaseTool._parse_input() extracts each field with getattr(model, alias). When an alias collides with a built-in pydantic model attribute (e.g. schema), this returns the built-in rather than the field value.
  2. tool_call_schema rebuilds a subset of the model field-by-field but drops alias and serialization options, so the rebuilt schema no longer matches what the LLM emits.

  Attachment aliases every field (id → ID, full_name → FullName, …), so it hits both bugs.

  Fix

  Introduce ExtractionToolInputSchema — an alias-free mirror of Attachment with the same python field names and annotations — and use it as the tool's args_schema. Sidesteps both LangChain bugs until they're fixed upstream.

  Changes

  - src/uipath_langchain/agent/tools/extraction_tool.py: define ExtractionToolInputSchema and use it as args_schema; docstring explains the reasoning so future readers don't switch back to Attachment.
  - tests/agent/tools/test_extraction_tool.py: replace the args_schema == Attachment assertion with a structural check that the schema's field names and annotations match Attachment.model_fields — guards against drift if
  Attachment evolves.

  Test plan

  - uv run pytest tests/agent/tools/test_extraction_tool.py passes
  - Verify end-to-end: deploy an agent that uses the IXP extraction tool and confirm download_async receives the populated UUID

### DIFF
--- a/src/uipath_langchain/agent/tools/extraction_tool.py
+++ b/src/uipath_langchain/agent/tools/extraction_tool.py
@@ -1,14 +1,15 @@
 """Ixp extraction tool."""
 
-from typing import Any
+import uuid
+from typing import Any, Optional
 
 from langchain.tools import BaseTool
 from langchain_core.messages import ToolCall, ToolMessage
 from langchain_core.tools import StructuredTool
 from langgraph.types import Command, interrupt
+from pydantic import BaseModel, Field
 from uipath.agent.models.agent import AgentIxpExtractionResourceConfig
 from uipath.eval.mocks import mockable
-from uipath.platform.attachments import Attachment
 from uipath.platform.common import DocumentExtraction
 from uipath.platform.documents import ExtractionResponseIXP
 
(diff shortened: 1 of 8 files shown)


src/uipath_langchain/agent/tools/extraction_tool.py
```diff
@@ -26,6 +27,34 @@ class StructuredToolWithWrapper(StructuredToolWithOutputType, ToolWrapperMixin):
...
+    aliased fields in two places (see PR #796):
+
+    1. `BaseTool._parse_input()` extracts each field with `getattr(model, key)`,
+       where `key` is the alias. For aliases that collide with built-in model
+       attributes (e.g. `schema`), this returns the built-in instead of the
+       field value, so downstream `kwargs.get("id") / kwargs.get("full_name")`
+       came back as `None`.
+    2. `tool_call_schema` rebuilds a subset of the model by copying each field
+       but drops alias and serialization options, so the rebuilt schema no
+       longer matches what the LLM emits.
```
andreitava-uip, line 44: nit: these bugs are not what causes it. It was us fixing the bugs via our structured tool subclass that surfaced the issue.

andreitava-uip (approved):

Issue #825, opened 2026-05-05
Breaking changes in authentication from uipath-langchain v0.9.24 -> v0.10.15?


mikael-andersson91:
For all of our UiPath LangChain agents, we are using the standard `UiPathChat` for our LLM models in agents. After upgrading from version 0.9.24, to version v0.10.15 of `uipath-langchain`, we are consistently getting errors such as this one when debugging our agents in local sessions:

```sh
HTTP Request: GET https://cloud.uipath.com/ingka/GlobalTest/agenthub_/llm/api/capabilities "HTTP/1.1 200 OK"
HTTP Request: POST https://cloud.uipath.com/ingka/GlobalTest/agenthub_/llm/api/chat/completions "HTTP/1.1 403 Forbidden"
❌ Error: Unexpected error occurred - Error: UiPathPermissionDeniedError: Forbidden (Status Code: 403) None
 
Traceback (most recent call last):
  File "C:\GIT\rpa-augmented-operations\uipath-coded-agents\ask-lattjo\.venv\Lib\site-packages\uipath\llm_client\utils\exceptions.py", line 254, in raise_for_status
    original_raise_for_status()
  File "C:\GIT\rpa-augmented-operations\uipath-coded-agents\ask-lattjo\.venv\Lib\site-packages\httpx\_models.py", line 829, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '403 Forbidden' for url '[https://cloud.uipath.com/ingka/GlobalTest/agenthub_/llm/api/chat/completions'](https://cloud.uipath.com/ingka/GlobalTest/agenthub_/llm/api/chat/completions%27)
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
```

Are there any breaking changes between these versions of uipath-langchain in terms of how to get our LLM calls authenticated that are not documented anywhere? This happens consistently when setting up new projects and running them according to the quickstart guide too. When we downgrade to older versions of the library, we do not get this error.

And follow-up question to that: Are there any plans from UiPaths side to use semantic versioning correctly to signal when breaking changes are made in releases for uipath-langchain?

radu-mocanu:
Thanks for the report.

We indeed introduced some big changes as part of our [v0.10.0](https://github.com/UiPath/uipath-langchain-python/releases/tag/v0.10.0) release around LLM models, but they should not have been breaking.

[...]

PR #834, opened 2026-05-08
fix: thread agenthub_config through chat_model_factory new path

cotovanu-cristian:
## Summary

When agents run with `command=\"debug\"` (or any non-runtime context), the LLM gateway calls were shipping `X-UiPath-AgentHub-Config: agentsruntime` instead of the computed `agentsplayground`. Every debug session was billed against the runtime quota.

## The bug

`chat_model_factory.get_chat_model` accepted an `agenthub_config` parameter whose docstring said it was \"ignored by the new factory\" — and it was. On the `use_new_llm_clients=True` path, the value was dropped before delegating to `uipath_langchain_client.factory.get_chat_model`, which built a fresh `PlatformSettings` whose `agenthub_config` defaults to `\"agentsruntime\"` upstream.

The local `_AgentHubConfigDefaultMixin` (added in #825/#830) clears that default for direct-construction users — `from uipath_langchain.chat import UiPathChat` etc. — but the factory path bypassed it entirely, so `uipath-agents` callers in `agent_graph_builder/llm_utils.py:create_llm` had their carefully computed value silently discarded.

Trace for `uipath debug agent.json`:
```
get_execution_type(ctx) → AgentExecutionType.PLAYGROUND
get_agenthub_config(PLAYGROUND) → \"agentsplayground\"  ✓
get_chat_model(model, agenthub_config=\"agentsplayground\")  → dropped on floor
get_chat_model_factory(client_settings=None)
  └─ PlatformSettings()  → agenthub_config=\"agentsruntime\"  ✗
```

## The fix

In `chat_model_factory.get_chat_model`, when `agenthub_config` is provided on the new path:
- Build `client_settings` via `get_default_client_settings()` if the caller didn't supply one.
- Set `client_settings.agenthub_config = agenthub_config` before delegating.
- Caller-supplied `client_settings` instances are **mutated**, not replaced, so other fields survive.
- `agenthub_config=None` is a pass-through — preserves existing env-var / direct-construction paths governed by the mixin.

`hasattr` guard mirrors the pattern in `_AgentHubConfigDefaultMixin._clear_agenthub_config_default`, since the `UiPathBaseSettings` ABC doesn't expose `agenthub_config` (only the concrete `PlatformBaseSettings` does).

## What changed

| File | Change |
|------|--------|
| `src/uipath_langchain/chat/chat_model_factory.py` | +12/-2 — fix + docstring update |
| `tests/chat/test_chat_model_factory_agenthub.py` | +123 — 4 new tests |
| `pyproject.toml` | 0.10.18 → 0.10.19 |
| `uv.lock` | regenerated |

## Tests added (4)

1. **Tracer** — `get_chat_model(agenthub_config=\"agentsplayground\", use_new_llm_clients=True)` results in `client_settings.agenthub_config == \"agentsplayground\"` reaching the upstream factory. *Was failing before the fix; proves the bug.*
2. **Pass-through** — `agenthub_config=None` does not synthesize a `client_settings`; preserves env-var / mixin paths.
3. **Caller settings preserved** — explicit `client_settings` is mutated, not replaced; non-`agenthub_config` fields survive.
4. **Legacy path regression guard** — `use_new_llm_clients=False` still threads `agenthub_config` to the legacy factory.

Existing `tests/chat/test_agenthub_config_default.py` (39 tests for direct-construction mixin behavior) still passes — the fix is additive, the mixin is still load-bearing for direct constructors.

Full chat suite: 239 passed.

## Out of scope / follow-ups

- **`ClaudeSDKRuntime` and `GatewayProxy` defaults** in `uipath-agents-python/src/uipath_agents/claude_sdk/{runtime.py:90,gateway_proxy.py:168}` — same `agenthub_config: str = \"agentsruntime\"` anti-pattern. Currently masked by `AgentsClaudeSDKRuntime` overriding both. Worth fixing but lives in a different repo/package.
- **Upstream `PlatformSettings.agenthub_config = \"agentsruntime\"` default** in `uipath_langchain_client` — would be a cleaner fix to set `default=None` upstream, but cross-package coordination. This PR neutralizes the symptom for the agents path without needing it.

## Test plan

- [x] New tests pass: `uv run pytest tests/chat/test_chat_model_factory_agenthub.py`
- [x] Existing mixin tests still pass: `uv run pytest tests/chat/test_agenthub_config_default.py`
- [x] Full chat suite passes: `uv run pytest tests/chat/` (239 passed)
- [x] mypy clean on changed source: `uv run mypy src/uipath_langchain/chat/chat_model_factory.py`
- [x] ruff clean on changed files
- [ ] Manual verification: run `uv run uipath debug agent.json '{}'` against a real tenant; confirm LLM gateway requests carry `X-UiPath-AgentHub-Config: agentsplayground`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-langchain"
-version = "0.10.18"
+version = "0.10.19"
 description = "Python SDK that enables developers to build and deploy LangGraph agents to the UiPath Cloud Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
--- a/src/uipath_langchain/chat/chat_model_factory.py
+++ b/src/uipath_langchain/chat/chat_model_factory.py
@@ -73,8 +73,10 @@ def get_chat_model(
             returned chat model. Accepts ``list[BaseCallbackHandler]`` or a
             ``BaseCallbackManager``. Forwarded only when explicitly set.
             Ignored by the legacy factory.
-        agenthub_config: AgentHub config header value. Required by the legacy
-            factory; ignored by the new factory.
+        agenthub_config: AgentHub config header value. Threaded onto
+            ``client_settings.agenthub_config`` on the new path (PlatformSettings
+            otherwise defaults to ``"agentsruntime"``); required by the legacy
+            factory.
         use_new_llm_clients: Routes to the new ``uipath_langchain_client``
             factory when True (default). When False, routes to the legacy
             in-repo clients.
@@ -94,6 +96,14 @@ def get_chat_model(
             byo_connection_id=byo_connection_id,
             **kwargs,
         )
+
+    if agenthub_config is not None:
+        if client_settings is None:
+            from uipath_langchain_client.settings import get_default_client_settings
+
+            client_settings = get_default_client_settings()
+        if hasattr(client_settings, "agenthub_config"):
+            client_settings.agenthub_config = agenthub_config
 
     optional_kwargs = {
         k: v
--- a/tests/chat/test_chat_model_factory_agenthub.py
+++ b/tests/chat/test_chat_model_factory_agenthub.py
@@ -1,0 +1,123 @@
+"""Tests for chat_model_factory.get_chat_model agenthub_config plumbing.
+
+The new-LLM-clients path must thread agenthub_config onto client_settings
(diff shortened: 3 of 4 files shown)


cotovanu-cristian:
closing as it's included in https://github.com/UiPath/uipath-langchain-python/pull/835

PR #856, opened 2026-05-18
feat: add client side tools to mapper and runtime [JAR-9629]

norman-le:
Had to make new PR (old one here: https://github.com/UiPath/uipath-langchain-python/pull/819)

```
Changes across repos for client-side tools feature:

uipath-langchain-python:
- client_side_tool.py: Factory for client-side tools. Creates tools with @durable_interrupt that pause the graph for client execution. Includes apply_tool_filter for names-only intersection filtering and available_client_side_tools ContextVar for per-exchange filtering. Tool function returns error ToolMessage if tool not in allowed set.
- hitl.py: Defines IS_CONVERSATIONAL_CLIENT_SIDE_TOOL constant. Confirmation interrupt sets is_execution_phase=True for server-side tools (bridge emits executingToolCall) and False for client-side tools (execution interrupt handles it separately).
- tool_node.py: Uses RunnableCallableWithTool for client-side tools (not just confirmation tools), enabling runtime discovery via .tool attribute.
- messages.py: MessageMapper emits startToolCall with isClientSideTool and outputSchema flags. Emits executingToolCall only for non-confirmed non-client-side server tools. Suppresses endToolCall for client-side tool results. Typed client_side_tools as dict[str, ClientSideToolInfo].
- runtime.py: Added _iter_graph_tools shared iterator. _get_client_side_tools returns dict[str, ClientSideToolInfo] discovered through RunnableCallableWithTool wrappers.
- agent.py: Derives client-side tools from the tools parameter (checking IS_CONVERSATIONAL_CLIENT_SIDE_TOOL metadata), scoped to is_conversational. Passes to create_init_node.
- init_node.py: Filters available client-side tools based on uipath__client_side_tools exchange input. None means all tools available. Non-list raises ValueError. Calls apply_tool_filter.

uipath-python:
- tool.py: ExecutingToolCallEvent has timestamp and input only (no toolName). Docstring says emitted in all scenarios.
- exchange.py: UiPathClientSideToolDeclaration model with name (required), input_schema (optional), output_schema (optional). UiPathConversationExchangeStartEvent has optional client_side_tools field.
- _bridge.py: emit_interrupt_event emits executingToolCall (without toolName) for triggers with is_execution_phase=True. Guards on tool_call_id only.

uipath-agents-python:
- graph.py: Strips IS_CONVERSATIONAL_CLIENT_SIDE_TOOL during evals.

AgentInterfaces:
- conversation-event.schemas.ts: ExecutingToolCallEventSchema has timestamp and input only (no toolName). Input documented as final input after possible confirmation modifications. ClientSideToolDeclarationSchema documented for filtering not validation. ExchangeStartEvent clientSideTools documented that omission means all tools supported.
- conversation_v1.service.ts: Removed agent package download validation (validateAndCreateExchange). clientSideTools passed through to runtime as uipath__client_side_tools (plain string names). Injection moved into getUnifiedRuntimeAgentJobStartInput before input size check/file offload.
- conversation-client-event.controller.ts: signalUserActivity accepts optional clientSideTools, passes through to service. Stores on exchange properties.
- ConversationProvider.tsx: Handles executingToolCall for client-side tools, uses toolName from startToolCall directly.
```

Videos in CAS localhost (Agent Builder changes already published, but this will be the debug experience as well when the sdk is upgraded):

Without tool confirmation for client side tool:

https://github.com/user-attachments/assets/48bc0979-0f51-471c-90aa-af561d876c31

With tool confirmation for client side tool:

https://github.com/user-attachments/assets/384ba36d-db2d-478e-8a74-8b3de0501a06

URT Eval:
image

Related to changes in other PRs:
CAS: https://github.com/UiPath/AgentInterfaces/pull/949
uipath-langchain-python (tool definition): https://github.com/UiPath/uipath-langchain-python/pull/856
uipath-python (bridge changes): https://github.com/UiPath/uipath-python/pull/1638
uipath-agents-python (skipping for evals, so it can be mocked): https://github.com/UiPath/uipath-agents-python/pull/485
Another Agents PR fixing some issues with debug: https://github.com/UiPath/Agents/pull/5250

### DIFF
--- a/src/uipath_langchain/agent/react/agent.py
+++ b/src/uipath_langchain/agent/react/agent.py
@@ -81,22 +81,22 @@ def create_agent(
     llm_tools: list[BaseTool] = [*agent_tools, *flow_control_tools]
 
     # Derive client-side tool schemas from tools for input validation in the init node.
-    cs_tools: dict[str, ClientSideToolInfo] | None = None
+    conversational_client_side_tools: dict[str, ClientSideToolInfo] | None = None
     if config.is_conversational:
-        cs_tools = {}
+        conversational_client_side_tools = {}
         for t in agent_tools:
             meta = getattr(t, "metadata", None) or {}
             if meta.get(IS_CONVERSATIONAL_CLIENT_SIDE_TOOL):
-                cs_tools[t.name] = {
+                conversational_client_side_tools[t.name] = {
                     "input_schema": t.args_schema.model_json_schema()
                     if hasattr(t, "args_schema") and t.args_schema
                     else None,
                     "output_schema": meta.get("output_schema"),
                 }
-        cs_tools = cs_tools or None
+        conversational_client_side_tools = conversational_client_side_tools or None
 
     init_node = create_init_node(
-        messages, input_schema, config.is_conversational, cs_tools
+        messages, input_schema, config.is_conversational, conversational_client_side_tools
     )
 
     tool_nodes = create_tool_node(agent_tools)
--- a/src/uipath_langchain/agent/react/agent.py
+++ b/src/uipath_langchain/agent/react/agent.py
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Sequence, Type, TypeVar
+from typing import Callable, Sequence, Type, TypeVar
 
 from langchain_core.language_models import BaseChatModel
 from langchain_core.messages import HumanMessage, SystemMessage
--- a/src/uipath_langchain/agent/react/agent.py
+++ b/src/uipath_langchain/agent/react/agent.py
@@ -1,4 +1,4 @@
-from typing import Callable, Sequence, Type, TypeVar
+from typing import Any, Callable, Sequence, Type, TypeVar
 
 from langchain_core.language_models import BaseChatModel
 from langchain_core.messages import HumanMessage, SystemMessage
@@ -8,6 +8,9 @@ from langgraph.graph import StateGraph
 from pydantic import BaseModel
 from uipath.platform.context_grounding import DeepRagContent
 from uipath.platform.guardrails import BaseGuardrail
+
+from uipath_langchain.agent.tools.client_side_tool import ClientSideToolInfo
+from uipath_langchain.chat.hitl import IS_CONVERSATIONAL_CLIENT_SIDE_TOOL
 
 from ...runtime._citations import cas_deep_rag_citation_wrapper
 from ..guardrails.actions import GuardrailAction
@@ -77,7 +80,24 @@ def create_agent(
     )
     llm_tools: list[BaseTool] = [*agent_tools, *flow_control_tools]
 
-    init_node = create_init_node(messages, input_schema, config.is_conversational)
+    # Derive client-side tool schemas from tools for input validation in the init node.
+    cs_tools: dict[str, ClientSideToolInfo] | None = None
+    if config.is_conversational:
+        cs_tools = {}
+        for t in agent_tools:
+            meta = getattr(t, "metadata", None) or {}
+            if meta.get(IS_CONVERSATIONAL_CLIENT_SIDE_TOOL):
+                cs_tools[t.name] = {
+                    "input_schema": t.args_schema.model_json_schema()
+                    if hasattr(t, "args_schema") and t.args_schema
+                    else None,
+                    "output_schema": meta.get("output_schema"),
+                }
+        cs_tools = cs_tools or None
+
+    init_node = create_init_node(
+        messages, input_schema, config.is_conversational, cs_tools
+    )
 
     tool_nodes = create_tool_node(agent_tools)
 
--- a/src/uipath_langchain/agent/react/agent.py
+++ b/src/uipath_langchain/agent/react/agent.py
@@ -89,14 +89,19 @@ def create_agent(
             if meta.get(IS_CONVERSATIONAL_CLIENT_SIDE_TOOL):
                 conversational_client_side_tools[t.name] = {
                     "input_schema": t.args_schema.model_json_schema()
-                    if hasattr(t, "args_schema") and t.args_schema
+                    if hasattr(t, "args_schema")
+                    and t.args_schema
+                    and hasattr(t.args_schema, "model_json_schema")
                     else None,
                     "output_schema": meta.get("output_schema"),
                 }
         conversational_client_side_tools = conversational_client_side_tools or None
 
     init_node = create_init_node(
-        messages, input_schema, config.is_conversational, conversational_client_side_tools
+        messages,
+        input_schema,
+        config.is_conversational,
+        conversational_client_side_tools,
     )
 
     tool_nodes = create_tool_node(agent_tools)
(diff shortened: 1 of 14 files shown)


src/uipath_langchain/agent/react/init_node.py
```diff
@@ -63,6 +71,19 @@ def graph_state_init(state: Any) -> Any:
             )
             job_attachments_dict.update(message_attachments)
 
+        # Validate client-side tool declarations from the exchange input
+        if client_side_tools:
+            client_tools_input = getattr(state, UIPATH_CLIENT_SIDE_TOOLS_INPUT_KEY, None)
+            if client_tools_input is None:
+                available_client_side_tools.set(None)
```
maxduu: Hm, didn't we agree to the following (based on Slack discussions / this [comment](https://github.com/UiPath/AgentInterfaces/pull/949#discussion_r3243364302))?

Given:
- Agent has client-tools from definition
- Client passes in `uipath__client_side_tools` names

We do:
1. If no `uipath__client_side_tools` passed in (it is None/null), then agent assumes to use all of their client-side-tools
2. If `uipath__client_side_tools` is an array of names, then filter out the agent's tools to only be the names in `uipath__client_side_tools`. If `uipath__client_side_tools` contains irrelevant/wrong names, that's okay for now I think.

So for example if agent has definition with tools: [A, B]
- Client passes in None/null `uipath__client_side_tools`: Agent has access to [A, B]
- Client passes in []: Agent filters out and has access to no client-side tools ([])
- Client passes in [A]: Agent filters out and has access to just [A]
- Client passes in [A, B]: Agent filters out and has access to [A, B]
- Client passes in [A, C, B]: Agent has access to [A, B]


Also, `validate_and_apply_tool_filter` seems to assume the `uipath__client_side_tools` is a list of tool-definitions rather than a list of strings, which you're passing from CAS?

src/uipath_langchain/runtime/messages.py
```diff
@@ -443,15 +448,39 @@ async def map_current_message_to_start_tool_call_events(self):
...
+                                UiPathConversationMessageEvent(
+                                    message_id=self.current_message.id,
+                                    tool_call=UiPathConversationToolCallEvent(
+                                        tool_call_id=tool_call["id"],
+                                        executing=UiPathConversationExecutingToolCallEvent(
+                                            input=tool_call["args"],
+                                        ),
+                                    ),
+                                )
+                            )
```
maxduu: **Post-sync summary:**

The core timing issues before were resulting from us emitting the `executingToolCall` when _firing_ the interrupts, not when the interrupt for the wait-for-confirmation resumes. It looks like we still need 2 codepaths due to the issue of emitting events at the tool-call graph node layer in `uipath_langchain`, but overall proposal that should be cleaner is:

1. Change this code in the mapper to
```
# Emit executingToolCall from MessageMapper for tools without tool-call confirmation
# Tools with tool-call confirmation get it emitted when the tool-call confirmation resumes.
if not require_confirmation:
      events.append(
         UiPathConversationMessageEvent(
             message_id=self.current_message.id,
             tool_call=UiPathConversationToolCallEvent(
             tool_call_id=tool_call["id"],
             executing=UiPathConversationExecutingToolCallEvent(
                 input=tool_call["args"],
       ),
    ),
  )
)
```

2. Handle tool-call-confirmation-tools at the **resume** step at: https://github.com/UiPath/uipath-runtime-python/blob/main/src/uipath/runtime/chat/runtime.py#L106-L110
- Check if tool-confirmation event returning, and if so, emit `executingToolCallEvent`
- The check is important so we don't emit `executingToolCallEvent` when the resume-value for the actual client-side-tool is obtained.

**Scenarios:**
1. Server side, no confirmation: MessageMapper emits executingToolCall, tool runs on server
2. Server, with confirmation: Interrupt for confirmation runs, and on resume, bridge emits executingToolCall. Tool runs.
3. Client-side, no confirmation: MessageMapper emits executingToolCall, tool runs on server
4. Client-side, with confirmation:  Interrupt for confirmation runs, and on resume, bridge emits executingToolCall. We do another wait for client to execute and send result.

src/uipath_langchain/runtime/messages.py
```diff
@@ -443,15 +448,39 @@ async def map_current_message_to_start_tool_call_events(self):
...
```
maxduu, line 482: Yeah this actually makes a ton of sense seeing it in code - thanks for the documentation comment as well. If it requires a confirmation, we send the event when the confirmation approval is received, otherwise directly here.

src/uipath_langchain/agent/react/agent.py
```diff
@@ -77,7 +80,24 @@ def create_agent(
     )
     llm_tools: list[BaseTool] = [*agent_tools, *flow_control_tools]
 
-    init_node = create_init_node(messages, input_schema, config.is_conversational)
+    # Derive client-side tool schemas from tools for input validation in the init node.
+    cs_tools: dict[str, ClientSideToolInfo] | None = None
```
maxduu: nit: Let's name this variable `conversational_client_side_tools` since `cs_tools` seems a little bit harder to follow?

src/uipath_langchain/agent/react/agent.py
```diff
@@ -77,7 +80,24 @@ def create_agent(
...
+    if config.is_conversational:
+        cs_tools = {}
+        for t in agent_tools:
+            meta = getattr(t, "metadata", None) or {}
+            if meta.get(IS_CONVERSATIONAL_CLIENT_SIDE_TOOL):
+                cs_tools[t.name] = {
+                    "input_schema": t.args_schema.model_json_schema()
+                    if hasattr(t, "args_schema") and t.args_schema
+                    else None,
+                    "output_schema": meta.get("output_schema"),
```
maxduu, line 96: Just a question - how come `input_schema` is from `t.args_schema` and `output_schema` is from `meta`?

maxduu (approved):
Small nits/questions.

Also needs uipath-runtime bump after the uipath-runtime PR is merged

src/uipath_langchain/agent/tools/client_side_tool.py
```diff
@@ -0,0 +1,130 @@
...
+                return {
+                    "tool_call_id": tool_call_id,
+                    "tool_name": tool_name,
+                    "input": kwargs,
+                }
+
+            result = await wait_for_client_execution()
+            return result.get("output", result) if isinstance(result, dict) else result
+
+        result = await execute_tool()
```
maxduu, line 103: Doesn't the tool-call end event contain `output`, and `isError`, and `cancelled`?

Not too worried about the 3rd but we should allow clients to pass in `isError` with output being an error as well, and having that read/set on the `ToolMessage`

src/uipath_langchain/agent/react/agent.py
norman-le: The `StructuredTool` class in LangChain has a built-in `args_schema` field for defining what inputs a tool accepts. There's no equivalent built-in field for output schema. So when creating client-side tools, the input schema goes into `args_schema` (LangChain's standard field) and the output schema gets stashed in the `metadata` dict as a custom key. That's why in `agent.py`, input_schema is read from `t.args_schema` and output_schema is read from `meta.get("output_schema")` — they live in different places because LangChain only has first-class support for the input side.

PR #857, opened 2026-05-18
fix: reinsert citation tags on chat history replay [JAR-9909]

JoshParkSJ:
## Summary
- When assistant messages are replayed through the chat history, the cleaned text was sent to the LLM without its prior `<uip:cite/>` markup. The LLM then "forgot" that it had cited sources and would over- or under-cite on the next turn.
- Add `reconstruct_text_with_citations` (inverse of `extract_citations_from_text`) and call it from the chat message mapper before emitting text content blocks.
- Bump version 0.10.24 → 0.10.25.

Screenshot 2026-05-18 at 7 11 21 PM

https://uipath.atlassian.net/browse/JAR-9909

### DIFF
--- a/tests/runtime/test_chat_message_mapper.py
+++ b/tests/runtime/test_chat_message_mapper.py
@@ -12,6 +12,7 @@ from langchain_core.messages import (
     ToolMessage,
 )
 from uipath.core.chat import (
+    UiPathConversationCitation,
     UiPathConversationCitationSourceMedia,
     UiPathConversationCitationSourceUrl,
     UiPathConversationContentPart,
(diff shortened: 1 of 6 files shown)


tests/runtime/test_chat_message_mapper.py
```diff
@@ -799,6 +800,70 @@ def test_map_messages_assistant_role_produces_ai_message(self):
...
+                    content_part_id="part-1",
+                    mime_type="text/markdown",
+                    data=UiPathInlineValue(inline="First and second."),
+                    citations=[
+                        UiPathConversationCitation(
+                            citation_id="cit-1",
+                            created_at=TEST_TIMESTAMP,
+                            updated_at=TEST_TIMESTAMP,
+                            offset=0,
+                            length=5,
```
maxduu: With the new citation format where we don't do any wrapping, shouldn't the length just be 0? Since the length determines the wrapping?

Like, wouldn't it be offset = 5, length = 0 here?

maxduu: Ohh, I see. So the `extract_citations_from_text` treats the offset to always start at 0 and go the length until next citation, e.g. 'First<uip:cite title="A" url="https://a.com" />' means offset = 0, length = 5.

This is actually technically incorrect compared to the [streamed format](https://github.com/UiPath/uipath-langchain-python/blob/main/src/uipath_langchain/runtime/_citations.py#L177) which returns the `start`/`end` fields in the same content-part event - representing an unwrapped citation. Basically it should be offset = x, length = 0 always.

So I feel that `extract_citations_from_text` and `reconstruct_text_with_citations` should probably reflect this? Might be an easy change? I know `extract_citations_from_text` is just used for the agent output, which isn't used for anything besides evals, so should be okay to change without affecting any actual runtime behavior.

JoshParkSJ: discussed async https://uipath-product.slack.com/archives/C090NH84361/p1779146224221209

tests/runtime/test_citations.py
```diff
@@ -757,6 +780,100 @@ def test_different_sources_get_different_numbers(self):
         assert citations[1].sources[0].number == 2
 
 
+class TestReconstructTextWithCitations:
+    """Test cases for reconstruct_text_with_citations - the inverse of extract_citations_from_text."""
+
+    def test_no_citations_returns_text_unchanged(self):
+        assert reconstruct_text_with_citations("hello world", []) == "hello world"
+        assert reconstruct_text_with_citations("hello world", None) == "hello world"
```
maxduu, line 788: Love the round-trip tests, thanks!

maxduu (approved):
Thanks!

PR #865, opened 2026-05-20
Add spans for escalation memory lookup

mjnovice:
## Summary
- wrap escalation memory cache search in custom OTel spans matching the episodic memory lookup shape
- attach escalation search request/response JSON, hit/miss state, matched count, and result data to the spans
- cover cache hit and cache miss span attributes in escalation memory tests

## Tests
- .venv/bin/ruff format src/uipath_langchain/agent/tools/escalation_memory.py tests/agent/tools/test_escalation_memory.py
- .venv/bin/ruff check src/uipath_langchain/agent/tools/escalation_memory.py tests/agent/tools/test_escalation_memory.py
- .venv/bin/pytest tests/agent/tools/test_escalation_memory.py

### DIFF
--- a/src/uipath_langchain/agent/tools/escalation_memory.py
+++ b/src/uipath_langchain/agent/tools/escalation_memory.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+from contextlib import contextmanager
 from typing import Any
 from uuid import UUID
 
@@ -29,9 +30,16 @@ logger = logging.getLogger(__name__)
 
 MEMORY_CACHE_HIT_METRIC = "MemoryCacheHit"
 MEMORY_CACHE_MISS_METRIC = "MemoryCacheMiss"
+ESCALATION_MEMORY_STRATEGY = "EscalationMemoryCache"
 
 _metric_counters: dict[str, Any] = {}
 _MISSING_VALUE = object()
+
+
+@contextmanager
+def _noop_context():
+    """No-op context manager when OTel is unavailable."""
+    yield None
 
 
 class EscalationMemoryFieldSetting(BaseModel):
@@ -81,11 +89,13 @@ class EscalationMemoryRetriever:
         self,
         memory_space_id: str,
         *,
+        memory_space_name: str | None = None,
         folder_path: str | None = None,
         memory_settings: EscalationMemorySettings | None = None,
         uipath_sdk: UiPath | None = None,
     ) -> None:
         self.memory_space_id = memory_space_id
+        self.memory_space_name = memory_space_name or ""
         self.folder_path = folder_path
         self.memory_settings = memory_settings or EscalationMemorySettings()
         self._uipath_sdk = uipath_sdk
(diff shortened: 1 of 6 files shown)


src/uipath_langchain/agent/tools/escalation_memory.py
```diff
@@ -97,16 +104,108 @@ async def aretrieve(
...
-            response = await self._raw_escalation_search(sdk, request)
+            if tracer
+            else _noop_context()
+        )
 
-        return _cached_result_from_search_response(response)
+        with lookup_span_ctx as lookup_span:
+            fewshot_span_ctx = (
+                tracer.start_as_current_span(
+                    "Apply dynamic few shot",
```
smflorentino: the strategy isn't technically dynamic few shot right? we bypass the escalation if the threshold matches?

src/uipath_langchain/agent/tools/escalation_memory.py
```diff
@@ -97,16 +104,108 @@ async def aretrieve(
...
+                    attributes={
+                        "openinference.span.kind": "applyDynamicFewShot",
+                        "type": "applyDynamicFewShot",
+                        "span_type": "applyDynamicFewShot",
+                        "uipath.custom_instrumentation": True,
+                        "memorySpaceName": "",
```
smflorentino: why isn't the name populated?

src/uipath_langchain/agent/tools/escalation_memory.py
```diff
@@ -97,16 +104,108 @@ async def aretrieve(
...
+        lookup_span_ctx = (
+            tracer.start_as_current_span(
+                "Find previous memories",
+                attributes={
+                    "openinference.span.kind": "agentMemoryLookup",
+                    "type": "agentMemoryLookup",
+                    "span_type": "agentMemoryLookup",
+                    "uipath.custom_instrumentation": True,
+                    "memorySpaceName": "",
+                    "memorySpaceId": self.memory_space_id,
```
smflorentino, line 134: name?

src/uipath_langchain/agent/tools/escalation_memory.py
```diff
@@ -537,6 +636,30 @@ def _record_custom_metric(metric_name: str, memory_space_id: str) -> None:
...
+        return payload
+
+    results = _read_value(payload, "results")
+    if results is not None:
+        return {"results": results}
+
+    return payload
+
+
+def _safe_len(value: Any) -> int:
```
smflorentino, line 684: ?

src/uipath_langchain/agent/tools/escalation_memory.py
mjnovice: Good catch. Episodic memory sets `memorySpaceName` from `memory_config.memory_space_name or ""`. I updated escalation memory to follow the same pattern: the retriever now accepts `memory_space_name`, and `create_escalation_tool` resolves it from the escalation memory resource or agent memory feature when available. It still falls back to empty string if only an ID is configured.

src/uipath_langchain/agent/tools/escalation_memory.py
mjnovice: Agreed. This is not dynamic few-shot behavior; it is escalation memory cache lookup/bypass. I kept the OpenInference span kind compatible with the existing memory exporter, but renamed the child span to `Apply escalation memory` and set `strategy=EscalationMemoryCache` on both memory spans.

src/uipath_langchain/agent/tools/escalation_memory.py
mjnovice: This was too generic. Episodic memory can directly call `response.model_dump(...)` because `search_async` returns the typed SDK model. Escalation has one extra case: we intentionally fall back to the raw response dict when the typed SDK validation rejects string answers from the current API. I replaced `_trace_payload` with `_serialize_search_response_for_trace`, which only does the typed `model_dump` when the response is a Pydantic model and otherwise serializes the raw fallback response as-is.

src/uipath_langchain/agent/tools/escalation_memory.py
```diff
@@ -97,16 +107,109 @@ async def aretrieve(
...
+            else _noop_context()
+        )
 
-        return _cached_result_from_search_response(response)
+        with lookup_span_ctx as lookup_span:
+            fewshot_span_ctx = (
+                tracer.start_as_current_span(
+                    "Apply escalation memory",
+                    attributes={
+                        "openinference.span.kind": "applyDynamicFewShot",
```
smflorentino, line 149: in a future PR, this need to be renamed as its not technically dynamic few shot. please work with the UI people to make sure these are renamed

src/uipath_langchain/agent/tools/escalation_memory.py
```diff
@@ -97,16 +107,109 @@ async def aretrieve(
...
         sdk = self._uipath_sdk if self._uipath_sdk is not None else UiPath()
+
+        results_count = 0
+        cached_result: EscalationMemoryCachedResult | None = None
         try:
-            response = await sdk.memory.escalation_search_async(
-                memory_space_id=self.memory_space_id,
-                request=request,
-                folder_path=self.folder_path,
+            from opentelemetry import trace as otel_trace
```
smflorentino, line 116: why isn't import at the top of the file?

src/uipath_langchain/agent/tools/escalation_memory.py
```diff
@@ -97,16 +107,109 @@ async def aretrieve(
...
+
+            with fewshot_span_ctx as fewshot_span:
+                try:
+                    try:
+                        response = await sdk.memory.escalation_search_async(
+                            memory_space_id=self.memory_space_id,
+                            request=request,
+                            folder_path=self.folder_path,
+                        )
+                    except ValidationError:
```
smflorentino, line 170: why do we have this fall back logic? explain with a comment

src/uipath_langchain/agent/tools/escalation_memory.py
```diff
@@ -97,16 +107,109 @@ async def aretrieve(
...
+                            ),
+                        )
+                        fewshot_span.set_attribute(
+                            "response",
+                            _serialize_search_response_for_trace(response),
+                        )
+                        fewshot_span.set_attribute(
+                            "fromMemory", cached_result is not None
+                        )
+                except Exception as error:
```
smflorentino, line 196: no test coverage for this

smflorentino (approved):

src/uipath_langchain/agent/tools/escalation_memory.py
mjnovice: Added `test_sets_memory_lookup_spans_to_error_on_search_failure`, which covers the search failure path and asserts both the lookup span and apply span are marked with error status.

src/uipath_langchain/agent/tools/escalation_memory.py
mjnovice: Added a comment explaining the fallback. The SDK response model can reject existing escalation memory rows where `answer` is stored as a JSON string, but the raw API payload is still usable and parsed below.

PR #868, opened 2026-05-20
feat(guardrails): add PRE+POST tool scope support for PII and HarmfulContent middleware [AL-445]

apetraru-uipath:
## What changed?

Added PRE+POST tool scope support for `UiPathPIIDetectionMiddleware` and `UiPathHarmfulContentMiddleware`, matching the behavior of UiPath low-code agents where guardrails are evaluated against **both** tool input (PRE) and tool output (POST).

Key changes:
- **`pii_detection.py` / `harmful_content.py`**: Added `stage: GuardrailExecutionStage = PRE_AND_POST` parameter. TOOL scope now evaluates the guardrail against tool args (PRE) and against the tool's return value (POST).
- **`_base.py`**: Extracted `_extract_tool_output_data`, `_extract_tool_input_data`, and `_create_tool_wrap_hook` into `BuiltInGuardrailMiddlewareMixin` to eliminate copy-paste duplication between the two middlewares. The inner closure was further extracted into `_run_tool_guardrail(request, handler)` to make the PRE+POST logic directly testable.
- **`samples/joke-agent/graph.py`**: Added `UiPathHarmfulContentMiddleware` on TOOL scope and explicit `stage=PRE_AND_POST` to both TOOL-scope middlewares.
- **Version bump**: `uipath-langchain` 0.11.3 → 0.11.4.

## How has this been tested?

- **27 new unit tests** (`tests/guardrails/middlewares/test_tool_wrap.py`): Cover all branches of `_extract_tool_input_data`, `_extract_tool_output_data`, and `_run_tool_guardrail` — tool-name skipping, PRE-only, POST-only, PRE+POST stage routing, `GuardrailBlockException` → `AgentRuntimeError` conversion, `AgentRuntimeError` re-raise, generic exception logging, request/result modification, and empty-output skipping.
- **4 new e2e parity test runs** (`test_guardrails_in_langgraph.py`): `test_tool_pii_post_block` and `test_tool_harmful_content_post_block`, each running against both middleware and decorator flavors, verifying that violations in tool output trigger `BlockAction` correctly.
- **Hook wiring tests** extended with TOOL-scope assertions for both affected middlewares.
- Validated end-to-end by running `samples/joke-agent` — HTTP logs show two guardrail API calls per tool invocation (PRE + POST).
- `ruff check` and `mypy` pass on all changed files.

- [ ] Under Feature Flag
- [x] None
- [ ] DB migrations required
- [ ] API removals or changes

### DIFF
--- a/src/uipath_langchain/chat/models.py
+++ b/src/uipath_langchain/chat/models.py
@@ -1,5 +1,13 @@
 import os
+from collections.abc import AsyncGenerator, Generator
+from typing import Any
 
+from langchain_core.callbacks import (
+    AsyncCallbackManagerForLLMRun,
+    CallbackManagerForLLMRun,
+)
+from langchain_core.messages import BaseMessage
+from langchain_core.outputs import ChatGenerationChunk
 from uipath_langchain_client.clients.normalized.chat_models import (
     UiPathChat as _UpstreamUiPathChat,
 )
@@ -13,8 +21,58 @@ def _default_factory() -> str:
     return os.getenv("UIPATH_MODEL_NAME", DEFAULT_MODEL_NAME)
 
 
+def _strip_chunk_metadata(chunk: ChatGenerationChunk, final_seen: bool) -> bool:
+    """Strip metadata fields that accumulate via string concatenation in AIMessageChunk.__add__.
+
+    Returns updated final_seen flag.
+    """
+    gi = chunk.generation_info
+    if not gi:
+        return final_seen
+
+    has_finish = bool(gi.get("finish_reason"))
+    if has_finish:
+        if final_seen:
+            # Duplicate final chunk (API sometimes sends finish_reason twice): strip all
+            for key in ("model_name", "id", "created", "finish_reason"):
+                gi.pop(key, None)
+        else:
+            final_seen = True
+    else:
+        # Intermediate chunk: strip metadata that string-concatenates across chunks
+        for key in ("model_name", "id", "created"):
+            gi.pop(key, None)
+    return final_seen
+
+
 class UiPathChat(_UpstreamUiPathChat):
-    pass
+    def _uipath_stream(
+        self,
+        messages: list[BaseMessage],
+        stop: list[str] | None = None,
+        run_manager: CallbackManagerForLLMRun | None = None,
+        **kwargs: Any,
+    ) -> Generator[ChatGenerationChunk, None, None]:
+        final_seen = False
+        for chunk in super()._uipath_stream(
+            messages, stop=stop, run_manager=run_manager, **kwargs
+        ):
+            final_seen = _strip_chunk_metadata(chunk, final_seen)
+            yield chunk
+
+    async def _uipath_astream(
+        self,
+        messages: list[BaseMessage],
+        stop: list[str] | None = None,
+        run_manager: AsyncCallbackManagerForLLMRun | None = None,
+        **kwargs: Any,
+    ) -> AsyncGenerator[ChatGenerationChunk, None]:
+        final_seen = False
+        async for chunk in super()._uipath_astream(
+            messages, stop=stop, run_manager=run_manager, **kwargs
+        ):
+            final_seen = _strip_chunk_metadata(chunk, final_seen)
+            yield chunk
 
 
 UiPathChat.model_fields["model_name"].default_factory = _default_factory
--- a/src/uipath_langchain/guardrails/middlewares/_base.py
+++ b/src/uipath_langchain/guardrails/middlewares/_base.py
@@ -5,7 +5,9 @@ import json
 import logging
 from typing import Any
 
+from langchain.agents.middleware import AgentMiddleware, wrap_tool_call
 from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, ToolMessage
+from langgraph.prebuilt.tool_node import ToolCallRequest
 from langgraph.types import Command
 from uipath.core.guardrails import (
     GuardrailValidationResult,
@@ -13,12 +15,19 @@ from uipath.core.guardrails import (
 )
 from uipath.platform import UiPath
 from uipath.platform.guardrails import BuiltInValidatorGuardrail
+from uipath.platform.guardrails.decorators import GuardrailExecutionStage
 from uipath.platform.guardrails.decorators._exceptions import GuardrailBlockException
 
 from uipath_langchain.agent.exceptions import AgentRuntimeError
 
 from ..models import GuardrailAction
-from ._utils import convert_block_exception, extract_text_from_messages
+from ._utils import (
+    convert_block_exception,
+    create_modified_tool_request,
+    create_modified_tool_result,
+    extract_text_from_messages,
+    sanitize_tool_name,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -35,6 +44,8 @@ class BuiltInGuardrailMiddlewareMixin:
     _guardrail: BuiltInValidatorGuardrail
     _name: str
     action: GuardrailAction
+    _tool_names: list[str] | None
+    _tool_stage: GuardrailExecutionStage
     _uipath: UiPath | None = None
 
     def _get_uipath(self) -> UiPath:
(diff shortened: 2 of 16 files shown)


src/uipath_langchain/guardrails/middlewares/_base.py
```diff
@@ -55,6 +69,121 @@ def _handle_validation_result(
...
+            return await handler(request)
+
+        input_data = self._extract_tool_input_data(request)
+
+        if self._tool_stage in (
+            GuardrailExecutionStage.PRE,
+            GuardrailExecutionStage.PRE_AND_POST,
+        ):
+            try:
+                result = self._evaluate_guardrail(input_data)
```
valentinabojan: `_evaluate_guardrail` (line 57) is a **synchronous** HTTP call to the UiPath guardrails API, but `_run_tool_guardrail` is `async`. With the new default of `PRE_AND_POST`, every tool invocation now blocks the event loop **twice** — once here and again on line 157. Under any concurrent agent load this is a real throughput regression vs. the old PRE-only path.

Cheapest fix is to push the sync call into a worker thread (needs `import asyncio` at the top):

```suggestion
                result = await asyncio.to_thread(self._evaluate_guardrail, input_data)
```

And the same swap on line 157. Long-term, an `async evaluate_guardrail` on the SDK side would be cleaner.

src/uipath_langchain/guardrails/middlewares/_base.py
```diff
@@ -55,6 +69,121 @@ def _handle_validation_result(
...
 
+    def _extract_tool_output_data(
+        self, result: ToolMessage | Command[Any]
+    ) -> dict[str, Any]:
+        """Extract tool output data from handler result for POST-stage evaluation."""
+        if isinstance(result, Command):
+            update = result.update if hasattr(result, "update") else {}
+            messages = update.get("messages", []) if isinstance(update, dict) else []
+            content = (
+                messages[0].content
```
valentinabojan: A `Command(update={"messages": [...]})` can carry multiple messages, and the `ToolMessage` isn't guaranteed to be first. If a flow emits e.g. `[AIMessage(...), ToolMessage(<real content>)]`, this branch reads `messages[0]`, sees it isn't a `ToolMessage`, sets `content = None`, and POST silently gets no content to validate.

Iterate to find the first `ToolMessage` instead:

```suggestion
            tool_msg = next(
                (m for m in messages if isinstance(m, ToolMessage)), None
            )
            content = tool_msg.content if tool_msg is not None else None
```

src/uipath_langchain/guardrails/middlewares/_base.py
```diff
@@ -32,6 +44,8 @@ class BuiltInGuardrailMiddlewareMixin:
     _guardrail: BuiltInValidatorGuardrail
     _name: str
     action: GuardrailAction
+    _tool_names: list[str] | None
+    _tool_stage: GuardrailExecutionStage
```
valentinabojan: `_tool_stage` and `_tool_names` are annotation-only here — no class-level binding. Today both subclasses (`pii_detection.py`, `harmful_content.py`) set them in `__init__`, so it works. But the moment a future TOOL-scope subclass forgets, `_run_tool_guardrail` will `AttributeError` at runtime — not at construction — and only when a specific tool is called. The mixin docstring also only mentions `_guardrail` / `_name` / `action` as required attrs, which makes this easy to miss.

```suggestion
    _tool_names: list[str] | None = None
    _tool_stage: GuardrailExecutionStage = GuardrailExecutionStage.PRE_AND_POST
```

And extend the docstring (lines 38–41) to list these alongside the others.

src/uipath_langchain/guardrails/middlewares/_base.py
```diff
@@ -55,6 +69,121 @@ def _handle_validation_result(
...
+    ) -> ToolMessage | Command[Any]:
+        """Execute PRE guardrail check, run tool handler, then POST guardrail check."""
+        tool_call = request.tool_call
+        tool_name = tool_call.get("name", "")
+        sanitized_tool_name = sanitize_tool_name(tool_name)
+
+        if self._tool_names is None or sanitized_tool_name not in self._tool_names:
+            return await handler(request)
+
+        input_data = self._extract_tool_input_data(request)
```
valentinabojan: `input_data` is extracted on every invocation, even when `stage=POST` and the value is never read. Cheap call, but pointless work — move it inside the PRE block (right after `try:` on line 132). Then a POST-only middleware skips the dict copy / `str()` conversion entirely.

src/uipath_langchain/guardrails/middlewares/_base.py
```diff
@@ -55,6 +69,121 @@ def _handle_validation_result(
...
+        if self._tool_stage in (
+            GuardrailExecutionStage.POST,
+            GuardrailExecutionStage.PRE_AND_POST,
+        ):
+            output_data = self._extract_tool_output_data(tool_result)
+            if output_data:
+                try:
+                    result = self._evaluate_guardrail(output_data)
+                    modified_output = self._handle_validation_result(result, output_data)
+                    if modified_output is not None:
```
valentinabojan, line 170: Asymmetry with line 135: PRE only re-wraps the request when `isinstance(modified_input, dict)`, but POST forwards any non-`None` to `create_modified_tool_result`. The helper actually handles both `str` and `dict` (`_utils.py:71-82`), so the difference is intentional — but it's load-bearing in a way that's not obvious from the call site, and the next refactor pass is likely to "tidy it up" and break things.

At minimum, add a one-line comment:

```suggestion
                    if modified_output is not None:
                        # POST: ToolMessage.content accepts both str and dict
                        tool_result = create_modified_tool_result(tool_result, modified_output)
```

Or unify the contract by letting `create_modified_tool_request` accept `str` too, and gating both sites the same way.

valentinabojan (commented):
Took a closer look at the new `BuiltInGuardrailMiddlewareMixin._run_tool_guardrail`. The refactor and test coverage are solid. Posting 5 inline comments on `_base.py`:

- Sync HTTP call from `async` context (now fires up to 2× per tool invocation under the new default)
- `_extract_tool_output_data` only inspects `messages[0]` of a `Command`
- `_tool_stage` / `_tool_names` are annotation-only on the mixin — AttributeError trap for future TOOL-scope subclasses
- `input_data` extracted unconditionally even on POST-only path (wasted work)
- PRE/POST asymmetry on the `isinstance(..., dict)` guard around `modified_*`

No approval blocker — flagging as `COMMENT`.

src/uipath_langchain/chat/models.py
```diff
@@ -13,8 +21,58 @@ def _default_factory() -> str:
...
+        else:
+            final_seen = True
+    else:
+        # Intermediate chunk: strip metadata that string-concatenates across chunks
+        for key in ("model_name", "id", "created"):
+            gi.pop(key, None)
+    return final_seen
+
+
 class UiPathChat(_UpstreamUiPathChat):
```
apetraru-uipath, line 48: The original UiPathChat was a no-op stub (pass). The change was needed to fix a real bug in the joke agent output.

  Root cause: When LangChain tracing callbacks are registered, ainvoke uses the streaming API path. Each SSE chunk from the UiPath
  LLM gateway includes model_name, id, and created in its generation_info. LangChain accumulates chunks via AIMessageChunk.__add__ →
   merge_dicts, which string-concatenates values for these keys. With 14 SSE chunks, the final AI message had:
  
  response_metadata.model_name = "gpt-4o-2024-08-06gpt-4o-2024-08-06gpt-4o-2024-08-06..."  # × 14
  response_metadata.finish_reason = "stopstop"  # sent in 2 final chunks

  The fix strips model_name, id, created from intermediate chunks (they only belong on the final chunk), and deduplicates when the
  API sends finish_reason twice. The parent class _UpstreamUiPathChat includes these fields in every chunk by design (for standalone
   streaming use), so the override is the correct layer to fix this without patching the upstream client.
  
  The change is minimal: a module-level _strip_chunk_metadata helper + overrides of the two streaming methods that call super() and
  filter each chunk through it. No behavior changes for non-streaming invocations.

valentinabojan (approved):

PR #871, opened 2026-05-21
feat: add arrayBuilder static arg support

andreiancuta-uipath:
## Summary

Adds support for the `arrayBuilder` static-argument variant. ArrayBuilder declares an array-typed parameter whose elements are addressed by indexed sibling entries (`$['items'][0]`, `$['items'][1]`, ...). At resolution time those indexed children are collected and assembled into a list in index order; gaps and out-of-scope nested paths become `None`. Child variants supported: `static`, `argument` (from agent input), and `textBuilder`.

### Example

Input properties:
```
"$['items']":    arrayBuilder
"$['items'][0]": static("alpha")
"$['items'][1]": argument(userId)        # agent_input = {"userId": "u-42"}
"$['items'][2]": textBuilder("Hi, {name}")  # agent_input = {"name": "world"}
```

Resolved value:
```
$['items'] -> ["alpha", "u-42", "Hi, world"]
```

### display_value

`ToolStaticArgument` now carries a separate `display_value` alongside `value`. The runtime call still receives the real `value`, but the schema modification step uses `display_value`, so sensitive arrayBuilder items render as `"<hidden>"` in the schema enum (e.g. `["<hidden>", "public"]`) while the tool itself is invoked with the real secret.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,11 @@
 [project]
 name = "uipath-langchain"
-version = "0.11.5"
+version = "0.11.6"
 description = "Python SDK that enables developers to build and deploy LangGraph agents to the UiPath Cloud Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
 dependencies = [
-    "uipath>=2.10.69, <2.11.0",
+    "uipath>=2.10.70, <2.11.0",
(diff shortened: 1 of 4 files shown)


andreiancuta-uipath:
@claude review

ionut-mihalache-uipath (approved):

PR #884, opened 2026-06-02
fix: do not add static args not found in the input schema

andreitava-uip:
For certain IS tools, there was a bug where certain IS fields were interpreted as static args despite not being in the schema. It has been fixed at design time, but agents with IS tools already configured won't get fixed without manually re-adding the tool.

Since we already had a warning for fields that could not be found in the schema, I extended to skip adding them to the tool input as well.

### DIFF
--- a/src/uipath_langchain/agent/tools/static_args.py
+++ b/src/uipath_langchain/agent/tools/static_args.py
@@ -164,23 +164,30 @@ ToolT = TypeVar("ToolT", bound=StructuredTool)
 def _apply_static_arguments_to_schema(
     tool: ToolT,
     static_args: dict[str, ToolStaticArgument],
-) -> ToolT:
+) -> tuple[ToolT, set[str]]:
     """Modify tool schema based on pre-resolved static arguments.
 
     Args:
         tool: The tool to modify
         static_args: The mapping from JSON paths to static arguments
+
+    Returns:
+        The schema-modified tool and the set of json paths that were applied to
+        the schema. Paths that cannot be applied are skipped, so callers must
+        not thread them into the tool call: the synthesized schema rejects them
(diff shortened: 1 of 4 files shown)


src/uipath_langchain/agent/tools/static_args.py
```diff
@@ -287,20 +295,30 @@ def initialize(
...
                 )
+                modified_tool, applied_paths = _apply_static_arguments_to_schema(
+                    tool, static_args
+                )
+                # Only thread args that survived schema modification: paths the
+                # schema rejected would fail the synthesized strict validator.
+                static_args = {
+                    path: sa
+                    for path, sa in static_args.items()
+                    if path in applied_paths
```
andreiancuta-uipath, line 313: You can move the `if` inside the `sanitize_dict_for_serialization` call so that you don't traverse the dict twice. `static_args` isn't used anywhere else.

And for clarity, you can extract the dict comprehension ouf of the function call to an `applied_static_values`

src/uipath_langchain/agent/tools/static_args.py
```diff
@@ -164,23 +164,30 @@ def deduplicate_argument_properties(
...
 
     Args:
         tool: The tool to modify
         static_args: The mapping from JSON paths to static arguments
+
+    Returns:
+        The schema-modified tool and the set of json paths that were applied to
+        the schema. Paths that cannot be applied are skipped, so callers must
+        not thread them into the tool call: the synthesized schema rejects them
+        as extra inputs.
```
andreiancuta-uipath: nit: remove the instructions for the caller. This function has no knowledge of how the args are stored later

andreiancuta-uipath (approved):

PR #899, opened 2026-06-09
feat: Governance runtime implementation

viswa-uipath:
**Dependent on : https://github.com/UiPath/uipath-python/pull/1761 for uipath-core package**

## Summary

Adds a LangChain / LangGraph governance adapter to `uipath-langchain`. It lets UiPath governance evaluate what a LangChain chain or LangGraph agent does at the model and tool level, and block disallowed actions, without the agent author writing governance code. This package contains only the LangChain-specific bridge.

## What it does

- Detects LangChain `Runnable` and LangGraph `CompiledStateGraph` objects and wraps them with a governance proxy.
- Hooks the model and tool lifecycle via LangChain's callback system and maps each event to a governance check:
  | LangChain event | Governance hook |
  |---|---|
  | `on_llm_start` / `on_chat_model_start` | `BEFORE_MODEL` |
  | `on_llm_end` | `AFTER_MODEL` |
  | `on_tool_start` | `TOOL_CALL` |
  | `on_tool_end` | `AFTER_TOOL` |
- Enforces by letting a `GovernanceBlockException` (raised on a DENY decision) propagate to stop the model call or tool. Any other error in the hook is logged and swallowed, so a governance failure cannot break an otherwise-healthy run.

## What it does not do

- Does not fire agent-level boundaries (`BEFORE_AGENT` / `AFTER_AGENT`); those are owned by the governance host.
- Does not depend on platform, auth, transport, or runtime internals — only on the shared governance contracts.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-langchain==0.13.23.dev1008995166",

  # Any version from PR
  "uipath-langchain>=0.13.23.dev1008990000,<0.13.23.dev1009000000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath-langchain>=0.13.23.dev1008990000,<0.13.23.dev1009000000",
]
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,14 +1,14 @@
 [project]
 name = "uipath-langchain"
-version = "0.13.22"
+version = "0.13.23"
 description = "Python SDK that enables developers to build and deploy LangGraph agents to the UiPath Cloud Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
 dependencies = [
     "uipath>=2.12.7, <2.13.0",
-    "uipath-core>=0.5.20, <0.6.0",
+    "uipath-core>=0.5.28, <0.6.0",
     "uipath-platform>=0.1.91, <0.2.0",
-    "uipath-runtime>=0.11.4, <0.12.0",
+    "uipath-runtime>=0.11.7, <0.12.0",
     "langgraph>=1.1.8, <2.0.0",
     "langchain-core>=1.2.27, <2.0.0",
     "langgraph-checkpoint-sqlite>=3.0.3, <4.0.0",
--- a/src/uipath_langchain/governance/__init__.py
+++ b/src/uipath_langchain/governance/__init__.py
@@ -1,0 +1,16 @@
+"""Governance integration for ``uipath-langchain``.
+
+Exposes :class:`GovernanceCallbackHandler` — a LangChain callback
+handler that calls an :class:`~uipath.core.adapters.EvaluatorProtocol`
+on the model and tool lifecycle. Wired into a run by passing an
+``evaluator`` to :class:`UiPathLangGraphRuntimeFactory`; the factory
+builds the handler and hands it to the runtime through the existing
+``callbacks`` channel.
+
+Importing this module has no side effects: no adapter is registered,
+no global state is mutated.
+"""
+
+from .callbacks import GovernanceCallbackHandler
+
+__all__ = ["GovernanceCallbackHandler"]
--- a/src/uipath_langchain/governance/callbacks.py
+++ b/src/uipath_langchain/governance/callbacks.py
@@ -1,0 +1,393 @@
+"""LangChain governance callback handler.
+
+A :class:`langchain_core.callbacks.BaseCallbackHandler` that calls a
+framework-agnostic :class:`~uipath.core.adapters.EvaluatorProtocol`
+on the model and tool lifecycle.
+
+Wiring lives in :class:`UiPathLangGraphRuntimeFactory`: passing an
+``evaluator`` to ``new_runtime`` causes the factory to build this
+handler and hand it to :class:`UiPathLangGraphRuntime` through the
+existing ``callbacks`` constructor arg. No adapter registry, no global
+state, no import-time mutation.
+
+Intercepts:
+
+- ``on_llm_start`` / ``on_chat_model_start`` / ``on_llm_end`` → BEFORE_MODEL / AFTER_MODEL
+- ``on_tool_start`` / ``on_tool_end``                         → TOOL_CALL / AFTER_TOOL
+
+Chain-level boundaries (BEFORE_AGENT / AFTER_AGENT) are intentionally
+*not* fired from here — they are owned by the governance host that
+drives the agent. ``ignore_chain = True`` makes LangChain skip chain
+notifications entirely, avoiding duplicate boundary evaluations.
+
+Audit emission and enforcement (raising
+:class:`GovernanceBlockException` on DENY) are owned by the evaluator
+itself. This module just hooks the framework callbacks, extracts the
+data, and calls ``evaluator.evaluate_*``; block exceptions propagate,
+everything else is logged and swallowed so a governance bug never
+breaks an agent run.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any, Iterable
+
+from langchain_core.callbacks import BaseCallbackHandler
+from langchain_core.messages import BaseMessage
+from langchain_core.outputs import (
+    ChatGeneration,
+    ChatGenerationChunk,
+    Generation,
+    GenerationChunk,
+    LLMResult,
+)
+from uipath.core.adapters import EvaluatorProtocol
+from uipath.core.governance.exceptions import GovernanceBlockException
+
+GenerationLike = Generation | ChatGeneration | GenerationChunk | ChatGenerationChunk
+
+logger = logging.getLogger(__name__)
+
+# Cap on the text scanned per model hook, so a long history / runaway
+# response can't blow scan-time budgets.
+_BEFORE_MODEL_TEXT_CAP = 64000
+
+
+class GovernanceCallbackHandler(BaseCallbackHandler):
+    """LangChain callback handler that fires governance evaluation.
+
+    The evaluator owns audit emission and DENY-raising. Each ``on_*``
+    callback only extracts the relevant payload and calls the matching
+    ``evaluate_*`` method; :class:`GovernanceBlockException` is allowed
+    to propagate, anything else is logged and swallowed.
+    """
+
+    run_inline: bool = True
+    raise_error: bool = True
+    ignore_llm: bool = False
+    # Chain-level events (BEFORE_AGENT / AFTER_AGENT) are owned by the
+    # governance host, so this handler skips them to avoid duplicate
+    # boundary evaluations.
+    ignore_chain: bool = True
+    ignore_agent: bool = False
+    ignore_retriever: bool = True
+    ignore_retry: bool = True
+    ignore_chat_model: bool = False
+    ignore_custom_event: bool = True
+
+    def __init__(
+        self,
+        evaluator: EvaluatorProtocol,
+        agent_name: str,
+        session_id: str,
+    ) -> None:
+        self._evaluator = evaluator
+        self._agent_name = agent_name
+        self._session_id = session_id
+        # ``trace_id`` is intentionally NOT held here. Trace correlation
+        # is owned by the layer below: OTel-backed sinks read the live
+        # span via ``trace.get_current_span()`` on the caller's thread;
+        # HTTP-bound consumers resolve the canonical trace id at call
+        # time. The callback handler is env-free and just forwards
+        # extracted payload to the evaluator.
+        self._session_state: dict[str, Any] = {"tool_calls": 0, "llm_calls": 0}
+        # Tool name lookup keyed by LangChain ``run_id`` so ``on_tool_end``
+        # can report the actual tool name to AFTER_TOOL evaluation.
+        self._tool_runs: dict[str, str] = {}
+
+    # ----- LLM callbacks ---------------------------------------------------
+
+    def on_llm_start(
+        self,
+        serialized: dict[str, Any],
+        prompts: list[str],
+        **kwargs: Any,
+    ) -> None:
+        """Evaluate BEFORE_MODEL rules at LLM start (non-chat completion).
+
+        ``prompts`` is a LangChain *batch* of independent completion
+        requests (``BaseLLM.generate(prompts=[...])``), not turns of a
+        single conversation. Every prompt is evaluated so a governed
+        request cannot bypass BEFORE_MODEL merely by not being the last
+        item in the batch; any DENY propagates.
+        """
+        try:
+            self._session_state["llm_calls"] = (
+                self._session_state.get("llm_calls", 0) + 1
+            )
+            for prompt in prompts:
+                self._evaluator.evaluate_before_model(
+                    model_input=prompt[:_BEFORE_MODEL_TEXT_CAP],
+                    agent_name=self._agent_name,
+                    runtime_id=self._session_id,
+                )
+        except GovernanceBlockException:
+            raise
+        except Exception as e:
+            logger.warning("on_llm_start governance check failed (continuing): %s", e)
+
+    def on_chat_model_start(
+        self,
+        serialized: dict[str, Any],
+        messages: list[list[BaseMessage]],
+        **kwargs: Any,
+    ) -> None:
+        """Evaluate BEFORE_MODEL rules for chat models.
+
+        ``messages`` is LangChain's nested batch shape
+        ``list[list[BaseMessage]]``: the **outer** list is the batch of
+        independent requests, each **inner** list is one request's full
+        message history. Every batch item is evaluated so an earlier
+        independent request cannot bypass BEFORE_MODEL, but within each
+        item only the **latest message** is scanned — not the full
+        history. The LLM still receives the entire history (this
+        callback doesn't mutate ``messages``); scoping to the newest
+        message keeps a violation in an earlier turn from re-firing on
+        every later turn that still carries it for context. Any DENY
+        propagates.
+
+        List-of-blocks content (multimodal, function-call, tool_use,
+        extended thinking) is walked via :meth:`_extract_block_text` so
+        dict-syntax noise from ``str(list)`` doesn't leak into the
+        regex-scanned blob.
+        """
+        try:
+            self._session_state["llm_calls"] = (
+                self._session_state.get("llm_calls", 0) + 1
+            )
+            for message_list in messages:
+                self._evaluator.evaluate_before_model(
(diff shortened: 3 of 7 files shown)


src/uipath_langchain/governance/adapter.py
```diff
@@ -0,0 +1,580 @@
...
+        # LangChain Runnable
+        try:
+            from langchain_core.runnables import Runnable
+
+            if isinstance(agent, Runnable):
+                return True
+        except ImportError:
+            pass
+
+        # Duck-typed fallback: anything with invoke + ainvoke that isn't
```
radu-mocanu: remove this broad duck-typed fallback. a LangChain adapter should only claim LangChain/LangGraph objects, or an explicit marker/registration path. `invoke` + `ainvoke` is too generic and can attach LangChain governance to other frameworks that use the same method names.

src/uipath_langchain/governance/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+    Idempotent — safe to call multiple times.
+    """
+    registry = get_adapter_registry()
+    if any(a.name == "LangChain" for a in registry.get_all()):
+        return
+    registry.register(LangChainAdapter())
+    logger.debug("Registered uipath-langchain governance adapter")
+
+
+# Side-effect registration on module import.
```
radu-mocanu: remove side-effect registration on import. this package already declares the `uipath.governance.adapters` entry point, so adapter registration should happen through the governance adapter discovery path. importing this module should not mutate the global registry.

src/uipath_langchain/governance/callbacks.py
```diff
@@ -0,0 +1,580 @@
...
+graphs. Uses LangChain's callback system for deep hooks (model / tool
+events) plus a thin proxy that ensures the callback is wired into
+``invoke`` / ``ainvoke`` / ``stream`` / ``astream``.
+
+This adapter intercepts:
+
+- ``on_llm_start`` / ``on_chat_model_start`` / ``on_llm_end`` → BEFORE_MODEL / AFTER_MODEL
+- ``on_tool_start`` / ``on_tool_end``                         → TOOL_CALL / AFTER_TOOL
+
+Chain-level boundaries (BEFORE_AGENT / AFTER_AGENT) are intentionally
```
radu-mocanu, line 18: keep this package focused on the LangChain adapter. remove references that tie it to a specific runtime wrapper location. `uipath-langchain` should implement the framework-specific hook bridge; runtime/core should discover and call it, not the adapter documenting runtime internals.

src/uipath_langchain/governance/adapter.py
aditik0303: Removed

src/uipath_langchain/governance/__init__.py
aditik0303: Removed

src/uipath_langchain/governance/callbacks.py
aditik0303: Done

src/uipath_langchain/governance/adapter.py
```diff
@@ -0,0 +1,593 @@
...
+
+            if isinstance(agent, Runnable):
+                return True
+        except ImportError:
+            pass
+
+        # Only LangChain/LangGraph objects are claimed. No duck-typed
+        # fallback: matching on generic ``invoke`` / ``ainvoke`` would let
+        # this adapter attach to other frameworks that share those method
+        # names.
```
radu-mocanu: no need for this comment now. this just shows tech debt (what you just removed)

src/uipath_langchain/governance/adapter.py
```diff
@@ -0,0 +1,593 @@
...
+
+    def on_tool_error(self, error: Exception, **kwargs: Any) -> None:
+        # Tool errored out — on_tool_end will not fire. Pop the mapping
+        # so a session with many failing tool calls does not leak.
+        run_id = kwargs.get("run_id")
+        if run_id is not None:
+            self._tool_runs.pop(str(run_id), None)
+        logger.warning("Tool error in governed session %s: %s", self._session_id, error)
+
+    # Chain-level callbacks (on_chain_start / on_chain_end / on_chain_error)
```
radu-mocanu: remove this comment. the module docstring and the `ignore_chain` setting already make it clear that chain callbacks are not handled here.

general note for this file: remove comments when the code is self explanatory. keep comments concise only when they explain current intent or a non-obvious constraint, not removed behavior / debt history.

src/uipath_langchain/governance/callbacks.py
```diff
@@ -0,0 +1,593 @@
...
+    callback only extracts the relevant payload and calls the matching
+    ``evaluate_*`` method; :class:`GovernanceBlockException` is allowed
+    to propagate, anything else is logged and swallowed.
+    """
+
+    # LangChain callback-handler descriptors:
+    run_inline: bool = True
+    raise_error: bool = False
+    ignore_llm: bool = False
+    # Chain-level events (BEFORE_AGENT / AFTER_AGENT) are owned by the
```
radu-mocanu, line 69: shorten this comment to the invariant: chain-level events are owned by the governance host, so this handler skips them to avoid duplicate boundary evaluations. no need to mention absent methods / AttributeError noise.

src/uipath_langchain/governance/adapter.py
aditik0303: Done — dropped the duck-typed fallback; `can_handle` now only claims a LangGraph `CompiledStateGraph` or a `langchain_core` `Runnable`.

src/uipath_langchain/governance/__init__.py
aditik0303: Done — removed the import-time registration; the adapter is registered only through the `uipath.governance.adapters` entry-point discovery path. Importing the module no longer mutates the global registry.

src/uipath_langchain/governance/callbacks.py
aditik0303: Done — the docstring now refers to the generic "governance host" and no longer documents the runtime wrapper / its internals.

src/uipath_langchain/governance/adapter.py
aditik0303: Done removed.

src/uipath_langchain/governance/adapter.py
aditik0303: Done removed

src/uipath_langchain/governance/callbacks.py
aditik0303: Done

radu-mocanu (approved):

radu-mocanu (commented):
please squash into 1 commit and rebase on top of main. make sure to bump the uipath-langchain version and regenerate the uv.lock

src/uipath_langchain/runtime/factory.py
```diff
@@ -286,14 +306,21 @@ async def _create_runtime_instance(
         )
 
     async def new_runtime(
-        self, entrypoint: str, runtime_id: str, **kwargs
+        self,
+        entrypoint: str,
+        runtime_id: str,
+        evaluator: EvaluatorProtocol | None = None,
```
cristipufu: use **kwargs to pass the evaluator - this avoids breaking changes in the uipath-runtime layer

radu-mocanu (changes requested):
please fix https://github.com/UiPath/uipath-langchain-python/pull/899#discussion_r3480007073 and remove the test pypi index from pyproject

src/uipath_langchain/governance/callbacks.py
```diff
@@ -0,0 +1,389 @@
...
+    """LangChain callback handler that fires governance evaluation.
+
+    The evaluator owns audit emission and DENY-raising. Each ``on_*``
+    callback only extracts the relevant payload and calls the matching
+    ``evaluate_*`` method; :class:`GovernanceBlockException` is allowed
+    to propagate, anything else is logged and swallowed.
+    """
+
+    run_inline: bool = True
+    raise_error: bool = False
```
radu-mocanu: This must be `True` for governance DENY to block execution. LangChain catches callback exceptions and only re-raises them when `handler.raise_error` is true, so with the current value a `GovernanceBlockException` raised by the evaluator is logged by LangChain and the model/tool call continues.

Please add a focused test that exercises LangChain callback dispatch, not only direct handler calls. Minimal shape:

```python
from langchain_core.callbacks.manager import handle_event

handler = GovernanceCallbackHandler(evaluator, "agent", "session")
evaluator.evaluate_before_model.side_effect = GovernanceBlockException("blocked")

with pytest.raises(GovernanceBlockException):
    handle_event([handler], "on_llm_start", "ignore_llm", {}, ["deny me"])
```

The same behavior should hold for async callback dispatch as well.

pyproject.toml
```diff
@@ -6,8 +6,8 @@ readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
 dependencies = [
     "uipath>=2.11.14, <2.12.0",
-    "uipath-core>=0.5.20, <0.6.0",
-    "uipath-platform>=0.1.79, <0.2.0",
+    "uipath-core>=0.5.25, <0.6.0",
```
radu-mocanu: Please also bump the package version for this logic-changing PR. This adds runtime governance behavior to `uipath-langchain`, and these packages publish on merge, so keeping `0.13.12` makes the change ride an existing version.

src/uipath_langchain/governance/callbacks.py
viswa-uipath: We are in audit mode now. We dont want to raise exceptions now. We will do it once we go with enforce mode . Post the private preview

pyproject.toml
viswa-uipath: We have to keep this PR open till the runtime changes are merged and the version is released. Will update the version post that.

src/uipath_langchain/governance/callbacks.py
```diff
@@ -0,0 +1,389 @@
...
+            return ""
+        pieces = (
+            self._extract_generation_text(gen)
+            for gen_list in response.generations
+            for gen in gen_list
+        )
+        return self._join_within_cap(pieces, "")
+
+    @staticmethod
+    def _extract_generation_text(gen: Any) -> str:
```
radu-mocanu: Please type this against LangChain output models instead of `Any` plus `getattr`. `on_llm_end` receives an `LLMResult`, and `LLMResult.generations` is already typed as `Generation | ChatGeneration | GenerationChunk | ChatGenerationChunk`, so this helper can use those public framework types and `isinstance` checks.

Suggested shape:

```python
from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, Generation, GenerationChunk

GenerationLike = Generation | ChatGeneration | GenerationChunk | ChatGenerationChunk

def _extract_generation_text(gen: GenerationLike) -> str:
    if isinstance(gen, ChatGeneration | ChatGenerationChunk):
        content = gen.message.content
        ...
    return gen.text or ""
```

The provider block content may still need loose typing, but the function parameter itself should be as typed as the framework contract allows.

src/uipath_langchain/governance/callbacks.py
radu-mocanu: can we add a todo comment at least?

src/uipath_langchain/runtime/factory.py
```diff
@@ -23,12 +25,16 @@
...
 
 from uipath_langchain._tracing import _instrument_traceable_attributes
+from uipath_langchain.governance import GovernanceCallbackHandler
 from uipath_langchain.runtime.config import LangGraphConfig
 from uipath_langchain.runtime.errors import LangGraphErrorCode, LangGraphRuntimeError
 from uipath_langchain.runtime.graph import LangGraphLoader
 from uipath_langchain.runtime.runtime import UiPathLangGraphRuntime
 from uipath_langchain.runtime.storage import SqliteResumableStorage
 
+_AGENT_TYPE_CODED = "Coded"
```
radu-mocanu: Please align these labels with the governance metadata taxonomy. Runtime governance documents `uipath_coded` for coded agents and `langchain` for this adapter framework; `langgraph` is the project marker/runtime flavor, not the framework label the host should stamp on governance telemetry. Add a `get_settings()` test for this contract.

src/uipath_langchain/governance/callbacks.py
```diff
@@ -0,0 +1,388 @@
...
+    ) -> None:
+        """Evaluate BEFORE_MODEL rules at LLM start (non-chat completion)."""
+        try:
+            self._session_state["llm_calls"] = (
+                self._session_state.get("llm_calls", 0) + 1
+            )
+            # Take only the latest prompt. Re-scanning every prompt in a
+            # batched call would re-fire rules on prior turns' content
+            # that's still in the prompt for context.
+            model_input = (prompts[-1] if prompts else "")[:_BEFORE_MODEL_TEXT_CAP]
```
radu-mocanu: This only evaluates the last prompt in the LangChain batch, so earlier batched requests bypass BEFORE_MODEL entirely. Please evaluate every prompt in `prompts` and let any DENY propagate. Same issue applies to chat batches below: the outer `messages` list is a batch, not history.

src/uipath_langchain/governance/callbacks.py
```diff
@@ -0,0 +1,388 @@
...
+itself. This module just hooks the framework callbacks, extracts the
+data, and calls ``evaluator.evaluate_*``; block exceptions propagate,
+everything else is logged and swallowed so a governance bug never
+breaks an agent run.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any, Dict, Iterable
```
radu-mocanu: Please use built-in generics and the concrete LangChain callback types where they exist. For example `dict[str, Any]` instead of `Dict[str, Any]`, and `messages: list[list[BaseMessage]]` instead of `list[list[Any]]`. Keep `Any` only where the framework API leaves no better type, such as passthrough `**kwargs`.

src/uipath_langchain/runtime/factory.py
viswa-uipath: Existing code in CLI holds the value `LowCode` not `uipath_lowcode` (https://github.com/UiPath/uipath-python/pull/1778/changes#diff-96b97b35e085eb0aa1d36c38b527cb2100829ede75e3103685ac2ceee3ea1d82L71). Shall I update runtime metadata docstring to match `LowCode` and `Coded` ?

src/uipath_langchain/governance/callbacks.py
viswa-uipath: That was intended. As stated in the comments. Otherwise it will trigger violations on previous conversation context.

radu-mocanu: This is not about scanning previous conversation context. The outer list is LangChain batching: multiple independent requests in one callback. The current implementation only evaluates the last batch item, so an earlier independent request can bypass BEFORE_MODEL.

This fails on the current implementation:

```python
from types import SimpleNamespace
from unittest.mock import MagicMock

import pytest
from uipath.core.governance.exceptions import GovernanceBlockException
from uipath_langchain.governance import GovernanceCallbackHandler


def _block_only_blocked(**kwargs):
    if "blocked" in kwargs["model_input"]:
        raise GovernanceBlockException("blocked")


def test_on_llm_start_evaluates_each_prompt_in_batch():
    evaluator = MagicMock()
    evaluator.evaluate_before_model.side_effect = _block_only_blocked
    handler = GovernanceCallbackHandler(evaluator, "agent", "session")

    with pytest.raises(GovernanceBlockException):
        handler.on_llm_start({}, ["blocked independent request", "safe independent request"])


def test_on_chat_model_start_evaluates_each_batch_item_latest_message():
    evaluator = MagicMock()
    evaluator.evaluate_before_model.side_effect = _block_only_blocked
    handler = GovernanceCallbackHandler(evaluator, "agent", "session")

    with pytest.raises(GovernanceBlockException):
        handler.on_chat_model_start(
            {},
            [
                [SimpleNamespace(content="blocked independent chat request")],
                [SimpleNamespace(content="safe independent chat request")],
            ],
        )
```

The fix should still evaluate only the latest message inside each batch item, not the full history. It just needs to iterate every independent batch item.

src/uipath_langchain/runtime/factory.py
radu-mocanu: I would not change the runtime governance metadata docs to `Coded` / `LowCode`.

Those look like a separate CLI/AppInsights taxonomy. Governance metadata should keep the two dimensions separate:

- `agent_type`: the product/host surface, for example `uipath_coded`, `uipath_lowcode`, external, etc.
- `agent_framework`: the technical adapter/framework, for example `langchain`, `llamaindex`, `openai_agents`, etc.

For this package I would use `agent_type="uipath_coded"` and `agent_framework="langchain"`.

If low-code is implemented on top of LangChain internally, that should not become `low-code-langchain` here. Low-code should set its own `agent_type` from its own package/host boundary, and either leave `agent_framework` unknown or set a low-code-owned value if product needs that. The OSS LangChain adapter should only report itself as `langchain`.

radu-mocanu (approved):

PR #902, opened 2026-06-11
fix: detect job attachments independent of dynamic model build order

radu-mocanu:
## Summary
- job attachment fields are now detected reliably even when several dynamic schemas (e.g. an agent's input and output, or a tool's args) are built in the same process

## Why

attachment detection compared field types by object identity against a single process-wide module attribute. each schema build registered its own attachment class under that attribute, so the most recent build won and detection on every other schema silently returned nothing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-langchain"
-version = "0.11.19"
+version = "0.11.20"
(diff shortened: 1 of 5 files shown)


gauravsingh-rajput:
Could we add one regression through get_job_attachments as well, not only get_json_paths_by_type? That would prove the user-facing attachment extraction still works after another dynamic model with the same JobAttachment definition is built?

gauravsingh-rajput (approved):

PR #904, opened 2026-06-14
chore(mcp): upgrade mcp client to 1.27.2 (2025-11-25 protocol)

edis-uipath:
## Upgrade the MCP client to `mcp==1.27.2` (2025-11-25 protocol)

Bumps the `mcp` SDK from **1.26.0 → 1.27.2** (latest stable), whose `LATEST_PROTOCOL_VERSION` is **`2025-11-25`**. The MCP client negotiates the SDK's latest protocol version on `initialize`, so this is what lets UiPath MCP servers advertise long-running work to the agent.

### Why the fork stays
The forked streamable-HTTP transport (`agent/tools/mcp/streamable_http.py`) exists to add UiPath's **session-id save + provide** layer (`SessionInfo`/`SessionInfoFactory`), which lets a suspended agent resume onto the **same** MCP session. We keep it because:
- The official `client/streamable_http.py` is **byte-identical between 1.26.0 and 1.27.2** (`git diff --stat` is empty), so there's nothing to re-sync against this release.
- The official client still offers **no way to provide an initial session id** through its public `streamable_http_client()` entrypoint. So the provide-on-resume capability still has to live in our fork.

The fork reads the **negotiated** protocol version from the server's `InitializeResult` (it doesn't hardcode one), so the bump alone flips negotiation to `2025-11-25` — no transport change needed.

### Changes
- `pyproject.toml`: `mcp==1.26.0` → `mcp==1.27.2`; `uv.lock` regenerated.

### Testing
- `tests/agent/tools/test_mcp/` — **57 passed**.
- Full suite — green; `ruff check` / `ruff format` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #905, opened 2026-06-14
feat(mcp): suspend the agent job on a UiPath MCP task

edis-uipath:
## Suspend the agent job on a UiPath MCP task

Stacked on #904 (mcp 1.27.2). When an MCP tool advertises task support and the call starts a **UiPath Orchestrator job**, the agent now **suspends** until the child job completes — instead of blocking on `tasks/result` — reusing the exact suspend/resume path `process_tool` already uses.

### Flow
1. A task-augmentable tool (per `AgentMcpTool.execution.taskSupport` = `optional`/`required`) is invoked.
2. `McpClient.call_tool_as_task` sends a **task-augmented** `tools/call` (`params.task`) and parses the `CreateTaskResult` (the Python SDK has no client task helper, so it sends the raw request via `send_request`).
3. The result's `_meta` marks it as a UiPath job (`uipath.com/source=orchestrator`, `jobKey`, `folderKey`).
4. `@durable_interrupt` → `interrupt(WaitJobRaw(job=Job(key=jobKey, folder_key=folderKey), process_folder_key=folderKey))`. The runtime persists a `JOB_RAW` resume trigger, Orchestrator **suspends the parent job**, and on child completion resumes with the job's output.

Non-task tools are unchanged (synchronous `call_tool`, 404-retry).

### Where task support comes from
- **Cached** discovery: `AgentMcpTool.execution.taskSupport` (UiPath/uipath-python#1716).
- **Dynamic** discovery: mapped from the server `Tool.execution` into the same field.

### ⚠️ Cross-package dependencies (this PR is the top of a stack)
- **uipath-python #1716** — adds `AgentMcpTool.execution.taskSupport`. **CI here is red until that publishes** and the lock picks it up (the import resolves then). Bump the `uipath` floor to that release before merge.
- **AgentHubService** — stamps the task result `_meta` marker **and** must emit a `ttl` on the task object (the official Python `Task.ttl` is required, so the `CreateTaskResult` won't parse without it).

### Testing
- Detection (`optional`/`required` ⇒ task; `forbidden`/absent ⇒ not), execution mapping from a server `Tool`, `call_tool_as_task` request shape, and the suspend path (asserts `WaitJobRaw` carries the job/folder keys from `_meta` and resume returns the job output).
- Full `test_mcp` suite green (**67**); ruff + mypy clean. (Run locally against an editable uipath-python #1716.)

### DIFF
--- a/src/uipath_langchain/agent/tools/mcp/mcp_client.py
+++ b/src/uipath_langchain/agent/tools/mcp/mcp_client.py
@@ -15,7 +15,15 @@ from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStre
 from mcp import ClientSession
 from mcp.shared.exceptions import McpError
 from mcp.shared.message import SessionMessage
-from mcp.types import CallToolResult, ListToolsResult
+from mcp.types import (
+    CallToolRequest,
+    CallToolRequestParams,
+    CallToolResult,
+    ClientRequest,
+    CreateTaskResult,
+    ListToolsResult,
+    TaskMetadata,
+)
 from uipath._utils._ssl_context import get_httpx_client_kwargs
 from uipath.runtime.base import UiPathDisposableProtocol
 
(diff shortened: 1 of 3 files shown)


edis-uipath:
we will wait to see where tasks land before doing this

PR #911, opened 2026-06-16
feat(datafabric): standalone ontology tool grounded on OWL + R2RML

sankalp-uipath:
## What

Adds a standalone **Data Fabric ontology tool** for low-code agents. The agent selects **ontologies** (not entities); the tool derives the entities it may query from the ontology's **R2RML**, resolves their schemas, and runs the existing inner SQL sub-graph grounded on both the **OWL** (semantic schema) and the **R2RML** (ontology→table/column mapping).

An ontology context (`contextType: "datafabricontology"`) now becomes a tool on its own — no `entitySet` required in `agent.json`.

## How it works (first invocation, cached)

1. Fetch the ontology's **R2RML** (critical — it is the entity allow-list) and **OWL** (grounds the prompt) via `EntitiesService.get_ontology_file_async`.
2. **Parse R2RML** with `rdflib` → the closed `(entity_name, folder_path)` allow-list, reading one `rr:tableName` (via `rr:logicalTable`) + one `uipath:folderPath` per `rr:TriplesMap`.
3. **Resolve** the allow-list **concurrently** (`asyncio.gather`): distinct `folderPath → folder_key` (`folders.retrieve_key_async`), then `name → id` (`entities.retrieve_by_name_async`), and delegate to the SDK's public `entities.resolve_entity_set_async` to fetch the schemas and build the folder-scoped `EntitiesService` (routing each entity's queries to its folder). Using the public resolver means the tool builds nothing from SDK internals.
4. Build a **dedicated ontology system prompt** (OWL + R2RML + entity schemas) and compile the ontology `DataFabricGraph` — the inner agent still has a **single tool, `execute_sql`**.

Everything from the sub-graph down (`execute_sql` → `query_entity_records_async`) mirrors the entity tool.

## Layout — the entity tool is left untouched

All ontology code lives in its own **`datafabric_tool/ontology/`** subpackage. The entity tool's files are **byte-identical to `main`**: `datafabric_tool.py`, `datafabric_subgraph.py`, `datafabric_prompt_builder.py`, and the package `__init__.py` are unchanged by this PR. `context_tool.py` is the only shared file touched — it builds the standalone ontology tool for the ontology context (flag-gated) and imports the tool directly from `datafabric_tool.ontology`.

**New — `datafabric_tool/ontology/`**
- `ontology_r2rml.py` — `rdflib`-based parser: `parse_r2rml_entities()` → the `(entity_name, folder_path)` allow-list (parses Turtle as an RDF graph; `R2RMLParseError` on invalid Turtle or contract violations).
- `ontology_tool.py` — `resolve_ontology_entities()` (the concurrent resolver), `DataFabricOntologyQueryHandler`, `create_datafabric_ontology_tool()`; owns the `DATAFABRIC_ONTOLOGY_FF` flag constant.
- `ontology_subgraph.py` — a self-contained, **prompt-agnostic** duplicate of `DataFabricGraph` (takes a pre-built `system_prompt`), so the entity sub-graph stays identical to `main`.
- `ontology_prompt_builder.py` — the ontology tool's inner prompt (OWL + R2RML + entity schemas), self-contained.
- `ontology_fetcher.py` — `fetch_ontology_file` (raw content + media type; raises) + `fence_ontology_block`.

## Design decisions

- **Closed allow-list (Option A):** the entities the tool can touch are exactly those declared in the ontology's R2RML — the LLM can never widen scope. `execute_sql`'s blast radius = the resolved set.
- **`uipath:folderPath` (a folder *path*, not a GUID):** R2RML has no folder concept and is deployment-agnostic; `rr:tableName` stays a valid SQL table name. Folder identity is resolved through the trusted folder service, never from mapping content. (Authoring contract documented for the ontology-authoring skill.)
- **`rdflib` parser:** the R2RML is parsed as an RDF graph rather than scanned with regexes, so authoring variations (predicate order, whitespace, compact one-line maps, prefix/`@base` differences) are handled by a real Turtle parser. Validated against the live `california-schools` mapping fetched from Data Fabric.
- **Duplicated, isolated sub-graph:** `ontology_subgraph.py` is an intentional copy of `datafabric_subgraph.py` so the entity path is not modified. It is excluded from Sonar copy-paste detection (`sonar.cpd.exclusions`) rather than sharing code, to preserve that isolation.
- **No SDK change** for the core feature — all new code is in `uipath-langchain-python` over existing public `uipath-platform` methods.

## Feature flag

- Gated by **`DataFabricOntologyEnabled`** (default **off**), a single shared constant owned by the ontology subpackage.
- **Two guards (defense in depth):** the tool-factory entry (`context_tool` — off ⇒ no tool created, feature fully inert) and the handler's lazy init (re-checks before any OWL/R2RML fetch/parse/resolve). Off ⇒ the agent runs exactly as before; the entity tool is flag-independent.

## Security

- Entity/folder scope is bounded by the R2RML allow-list; folder keys come only from trusted folder-path resolution (not the LLM).
- `execute_sql`'s single-statement `sqlparse` guard is unchanged.
- OWL/R2RML are injected as prompt context; a hostile ontology can only reference entities it itself declares (Option A), which still pass folder-scoped resolution.

## Testing

- Unit tests for the parser (incl. the live `california-schools` shape), resolver (concurrent folder-key + by-name id fetch, delegation to `resolve_entity_set_async`), the ontology prompt builder, the ontology sub-graph nodes, the factory, and the flag guard. New ontology modules are covered ≥99%.
- **Live e2e on alpha** (`datafabric/ab`, gpt-5.4, `california-schools` ontology): flag prefetched on → R2RML + OWL fetched → `folderPath` resolved → 3 entities resolved by name → OWL+R2RML-grounded SQL (join on the R2RML FK) executed folder-scoped → correct answer. Confirmed `retrieve_by_name` (`/metadata`) returns populated field schemas.

## Notes / dependencies

- **Depends on SDK #1728** (merged), which ships the ontology binding: `ontologySet` on the context, `AgentContextType.DATA_FABRIC_ONTOLOGY`, `get_ontology_file_async`, and `resolve_entity_set_async`. The feature needs `uipath>=2.13.2` / `uipath-platform>=0.2.1`; `pyproject.toml` currently pins the higher `uipath>=2.13.5` / `uipath-platform>=0.2.4` adopted from `main`. No env-based configuration fallback is used — ontologies come only from the context's `ontologySet`.
- Adds **`rdflib>=7.0.0, <8.0.0`** for the R2RML parser.
- Runtime flag delivery also requires `DataFabricOntologyEnabled` in `uipath-agents-python`'s `_ALL_FLAGS` prefetch + the gitops flag deployed for the target tenants (both already in place from the prior work).
- **Authoring dependency:** ontologies must be published with a `uipath:folderPath` per `rr:TriplesMap` (authoring-skill guidelines provided) — otherwise resolution fails loudly by design.
- Supersedes the earlier "inject ontology into the entity tool" approach and folds in the R2RML grounding work.


src/uipath_langchain/agent/tools/datafabric_tool/ontology_client.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Raises:
+        ValueError: If the name is invalid or the content exceeds the size cap.
+            Transport/HTTP errors propagate from the SDK as raised exceptions
+            (the caller decides how to degrade).
+    """
+    safe_name = _validate_ontology_name(ontology_name)
+    # Same datafabric_ service the entities calls target; matches the
+    # QueryEngine ontology route GET /ontologies/{ontologyName}/files/{fileType}.
+    endpoint = f"datafabric_/api/ontologies/{safe_name}/files/owl"
```
milind-jain-uipath: these need to be stitched in uipath-python

milind-jain-uipath (changes requested):

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_subgraph.py
```diff
@@ -88,18 +89,29 @@ def __init__(
...
+        ontologies: list[tuple[str, str | None]] | None = None,
     ) -> None:
         self._max_iterations = max_iterations
         self._execute_sql_tool = self._create_execute_sql_tool(
             entities_service, entities
         )
+        # Inner toolset: always execute_sql; optionally an LLM-decided
+        # fetch_ontology tool when one or more ontologies are configured.
+        inner_tools: list[BaseTool] = [self._execute_sql_tool]
+        if ontologies:
```
UIPath-Harshit: EnabledNewLlmClients <- check for the feature flag impl of this to ensure out feature is behind the feature flag.

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_subgraph.py
```diff
@@ -88,18 +89,29 @@ def __init__(
...
+            inner_tools.append(
```
UIPath-Harshit: This doesnt update the subgraph ? correct?

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_subgraph.py
```diff
@@ -131,35 +143,62 @@ async def tool_node(self, state: DataFabricSubgraphState) -> dict[str, Any]:
             *[self._execute_tool_call(tc) for tc in last.tool_calls]
         )
         tool_messages = [msg for msg, _ in results]
-        all_succeeded = bool(results) and all(success for _, success in results)
+        # End as soon as ANY tool call is a terminal success (a row-returning
+        # execute_sql). `any` not `all`: a non-terminal tool (e.g. fetch_ontology)
+        # co-issued in the same turn must not prevent a successful SQL from ending
+        # the loop.
+        any_succeeded = any(success for _, success in results)
```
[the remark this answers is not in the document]
UIPath-Harshit: Instead of any_ check for FF to see what graph gets constructed.

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_tool.py
```diff
@@ -30,6 +30,19 @@
...
+def resolve_context_ontologies(
+    resource: AgentContextResourceConfig,
+) -> list[tuple[str, str | None]]:
+    """Map a context's nested ``ontology_set`` to ``(name, folder_key)`` pairs.
+
+    Ontologies are configured inline on the Data Fabric context (alongside the
+    entity set) as ``ontologySet`` items. Each carries its own ``folderId``, so
+    it is fetched from its own folder.
+    """
+    items = getattr(resource, "ontology_set", None) or []
```
UIPath-Harshit: Same as other PR. ontology_set?

sankalp-uipath: yes

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_subgraph.py
sankalp-uipath: yes, will add it.

tests/agent/tools/test_datafabric_ontology_subgraph.py
```diff
@@ -0,0 +1,154 @@
...
+    return {"name": name, "args": args or {}, "id": cid, "type": "tool_call"}
+
+
+def test_fetch_ontology_bound_only_when_ontologies(make_graph):
+    without = make_graph(None)
+    assert "execute_sql" in without._tools_by_name
+    assert "fetch_ontology" not in without._tools_by_name
+
+    with_onto = make_graph([("library", None)])
+    assert "fetch_ontology" in with_onto._tools_by_name
```
andreitava-uip: nit: splitting this test into two(should bind when present/should not bind when absent) is trivial and allows instantly knowing what failed from the test name alone without checking the assertion message.

src/uipath_langchain/agent/tools/datafabric_tool/ontology_fetch_tool.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+
+    def __init__(
+        self,
+        entities_service: EntitiesService,
+        ontologies: list[tuple[str, str | None]],
+    ) -> None:
+        self._entities_service = entities_service
+        self._ontologies = ontologies
+        self._cached: str | None = None
```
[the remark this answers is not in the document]
andreitava-uip: outer agent does not support true parallel invocation. Parallel tool calls are executed sequentially by the outer agent, so each SQL agent  instance won't be concurrently called.

src/uipath_langchain/agent/tools/context_tool.py
```diff
@@ -158,17 +158,27 @@ def create_context_tool(
 ) -> StructuredTool | BaseTool | None:
     tool_name = sanitize_tool_name(resource.name)
 
+    # An ontology context is not a standalone tool — it only grounds the Data
+    # Fabric entity tool, which gathers it via resolve_context_ontologies.
+    if resource.context_type == AgentContextType.DATA_FABRIC_ONTOLOGY:
+        return None
```
andreitava-uip: If it is not a standalone tool at runtime, I think it is confusing to model it as a top level resource at design time. So far, all "resource nodes" in a lowcode agent (either standalone or part of flow), are independently executable and show up in traces. This is now a different paradigm, it is an optional helper tool that will be part of another tool's subgraph.

That being said this only applies to how it's modeled today. If we indeed plan to expand ontology support in the future such that they will actually allow queries (via something like SPARQL statements for instance); then it will be better for future proofing to define them top level (at least in the package mapping). We can figure out a less confusing design time experience for now

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_prompt_builder.py
```diff
@@ -143,6 +151,26 @@ def format_sql_context(ctx: SQLContext) -> str:
...
+        )
+        lines.append("")
+        lines.append(
+            "**Before writing any SQL, call the `fetch_ontology` tool once** to "
+            "load it, then base your column names, filter values, and joins on "
+            "what it says. The entity tables below are a quick reference only; "
+            "the ontology is the source of truth when they disagree."
+        )
+        lines.append("")
+
```
andreitava-uip: nit: could be cleaner to have this as a single formatted string depending on names instead of individually applying each line like this.

Applicable to the existing `sql_expert_system_prompt` as well, but that one wasn't introduced by this PR

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_subgraph.py
```diff
@@ -130,36 +159,69 @@ async def tool_node(self, state: DataFabricSubgraphState) -> dict[str, Any]:
...
+        # When short-circuiting to END, return ONLY the terminal-success
+        # ToolMessages so the outer agent's result is the query rows — not a
+        # co-issued fetch_ontology's OWL. On a non-terminal turn keep all messages
+        # so the inner LLM can use them on its next pass.
```
andreitava-uip: isn't concurrent execution of a ontology retrieval + data service query an anomaly? It doesn't seem to be correct. Why not mechanically enforce ontology retrieval and injecting it in the context. When is it useful for the llm to choose *not* to fetch the ontology?

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_tool.py
```diff
@@ -29,6 +29,34 @@
...
+    """
+    ontologies: list[tuple[str, str | None]] = []
+    for resource in resources:
+        if (
+            isinstance(resource, AgentContextResourceConfig)
+            and resource.is_datafabric_ontology
+        ):
+            for item in resource.ontology_set or []:
+                ontologies.append((item.name, item.folder_key))
+    return ontologies
```
andreitava-uip: if I understand correctly, we implicitly assume all topologies will apply to this data service entity context. Shouldn't the link be more explicitly defined? IE either:
a) when defining an Data Service Context resource you can also specify one or more ontologies
b) when defining the Ontology Context resource you specify the list of entities it describes

sankalp-uipath: I am currently working on adding the R2RML mapping which will resolve the entities from ontologies at the agent runtime by the llm node implicitly (I am working on it in separate PR and is currently in progress ).

src/uipath_langchain/agent/tools/context_tool.py
sankalp-uipath: Yes, we have plan to expand ontology support to make it a primary design experience i.e, user will select the ontologies and then it will resolve the entities internally, thus decision of making it top level resource as a part of iterative development.

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_prompt_builder.py
sankalp-uipath: Fixed, please review. Also there are some changes linked to your other comment in the data fabric prompt builder (adding ontology text in the prompt).

src/uipath_langchain/agent/tools/datafabric_tool/datafabric_subgraph.py
sankalp-uipath: Agreed, earlier we were doing so as to support future use cases where llm will query the ontology (for ex using SPARQL), instead of giving the complete ontology to agent.
But I agree with you right now it makes more sense to mechanically injecting it in the system prompt.
I have made the changes please review again.

src/uipath_langchain/agent/tools/datafabric_tool/ontology/ontology_prompt_builder.py
```diff
@@ -0,0 +1,145 @@
...
+
+        lines.append(f"**Query Patterns for {entity.entity_name}:**")
+        lines.append("")
+        lines.append("| User Intent | SQL Pattern |")
+        lines.append("|-------------|-------------|")
+        for p in entity_ctx.query_patterns:
+            lines.append(f"| '{p.intent}' | `{p.sql}` |")
+        lines.append("")
+
+    return lines
```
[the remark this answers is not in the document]
sankalp-uipath: deliberate separation of concern.

src/uipath_langchain/agent/tools/datafabric_tool/ontology/ontology_r2rml.py
```diff
@@ -0,0 +1,83 @@
...
+The extracted ``(entity_name, folder_path)`` set is the closed allow-list of
+entities the ontology tool is permitted to resolve and query.
+"""
+
+from rdflib import Graph, Namespace
+from rdflib.namespace import RDF
+
+# R2RML vocabulary and the UiPath extension carrying the folder path.
+RR = Namespace("http://www.w3.org/ns/r2rml#")
+UIPATH = Namespace("http://uipath.com/ns/datafabric#")
```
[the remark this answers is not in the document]
sankalp-uipath: These are RDF namespace IRIs, opaque identifiers that must match the terms in the parsed Turtle verbatim, not network endpoints. The R2RML namespace is defined as http by the W3C spec; switching to https would stop the terms from matching.

src/uipath_langchain/agent/tools/datafabric_tool/ontology/ontology_tool.py
```diff
@@ -0,0 +1,312 @@
...
+from langchain_core.language_models import BaseChatModel
+from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
+from langchain_core.tools import BaseTool
+from langgraph.graph.state import CompiledStateGraph
+from uipath.agent.models.agent import AgentContextResourceConfig
+from uipath.platform.entities import DataFabricEntityItem, EntitiesService, Entity
+
+from ...base_uipath_structured_tool import BaseUiPathStructuredTool
+from ..datafabric_tool import BASE_SYSTEM_PROMPT
+from ..models import DataFabricQueryInput
```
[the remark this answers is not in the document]
sankalp-uipath: FALSE POSITIVE. from ..datafabric_tool import BASE_SYSTEM_PROMPT resolves to the sibling module datafabric_tool/datafabric_tool.py (which defines BASE_SYSTEM_PROMPT = "base_system_prompt"), not the package __init__.py. The import works (import OK -> DataFabricOntologyEnabled), and all tests + the circular-import test pass. Copilot confused the module with the same-named package. It's correct — but the ambiguity is real (it fooled the reviewer), so I'll add a clarifying comment.**

tests/agent/tools/test_ontology_fetcher.py
```diff
@@ -0,0 +1,85 @@
+"""Tests for ontology file fetching (datafabric_tool/ontology_fetcher.py)."""
```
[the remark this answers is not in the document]
sankalp-uipath: Fixed

UIPath-Harshit (commented):
## PR Review: feat(datafabric): standalone ontology tool grounded on OWL + R2RML

**Overall: Well-structured, thoughtful PR.** The isolation strategy (separate subpackage, entity tool untouched) is the right call. A few concerns below.

### Positives

1. **Clean isolation** — all ontology code in `datafabric_tool/ontology/`, entity tool byte-identical to main. Sonar CPD exclusion is the right trade-off for keeping them decoupled.
2. **Closed allow-list (Option A)** — entities bounded by R2RML, LLM can't widen scope. Good security posture.
3. **rdflib over regex** — real Turtle parser handles authoring variations correctly. Tests validate compact/multi-line/reordered syntax.
4. **Defense in depth** — flag checked at both tool creation (`context_tool.py`) and lazy init (`_ensure_graph`). Tool is fully inert when off.
5. **OWL degrades gracefully, R2RML is fatal** — correct priority since R2RML is the allow-list.
6. **Test coverage** — comprehensive unit tests for parser, resolver, subgraph, prompt builder, fetcher. Edge cases covered well.

### Concerns

**1. `rdflib` as a new dependency (medium)**
- `rdflib` pulls in `pyparsing`. This adds ~5MB to the package for a feature-flagged path. Is lazy importing the `rdflib` module an option? Currently `ontology_r2rml.py` imports it at module level — if `context_tool.py` does the lazy import of the ontology package only when the flag is on, this is probably fine, but worth confirming `rdflib` doesn't get loaded for non-ontology agents.

**2. Duplicated sub-graph (`ontology_subgraph.py` — 233 lines)**
- I understand the rationale (entity tool stays untouched), but this is a full copy of `DataFabricGraph` + `QueryExecutor` + state model. The only difference is the constructor takes `system_prompt: str` instead of building it internally. Could the entity sub-graph be refactored to also accept a pre-built prompt, making duplication unnecessary? The PR description says "deliberately isolated" — if this is a temporary measure until the entity tool is also refactored, document the intended convergence timeline.

**3. Folder resolution is sequential-then-concurrent but not batched optimally (`ontology_tool.py:725-753`)**
- Step 1a resolves distinct folder paths concurrently (good). Step 1b resolves entities by name concurrently (good). But steps 1a and 1b are sequential — 1b waits for all folder keys before starting any entity lookups. Since entity lookups need the folder key, this is correct, but worth noting that for N entities across M folders, latency = `folder_resolve_time + entity_resolve_time` (not parallelized across steps). Acceptable for small ontologies, but flag if california-schools (3 entities, 1 folder) takes noticeable time.

**4. `_MAX_ONTOLOGY_BYTES = 2_000_000` (`ontology_fetcher.py:114`)**
- 2MB of raw text injected into a prompt is a lot of tokens (~500K+). The size cap protects against blowup, but in practice even a 200KB OWL could degrade LLM quality. Consider a tighter warning threshold or token-budget-aware truncation, especially if customers publish large ontologies.

**5. `asyncio.Lock` for lazy init (`ontology_tool.py:778`)**
- This is a per-handler lock, so it's scoped correctly. But if the handler is shared across event loops (unlikely but possible in some deployment patterns), `asyncio.Lock` would fail. Just a note — probably fine for the current runtime.

**6. Missing error context in `resolve_ontology_entities` (`ontology_tool.py:742-746`)**
- When `retrieve_by_name_async` fails for one entity in the `asyncio.gather`, the whole batch fails with the first exception. Consider `return_exceptions=True` + a summary error that names which entities failed, to aid debugging multi-entity ontologies.

**7. Test: `test_factory_builds_handler_from_ontology_set` accesses private `_ontologies` (`test_datafabric_ontology_tool.py:1321`)**
- Minor, but fragile. Consider testing via the tool description or metadata instead.

### Nits

- `ontology_r2rml.py:344-345`: The `# NOSONAR` comments are clear, but consider adding a `ruff: noqa` equivalent if ruff also flags `http://` in strings.
- `ontology_prompt_builder.py:199`: `(field.description or "").replace("|", r"\|").replace("\n", " ")` — good defensive escaping. The entity tool's prompt builder should probably do the same if it doesn't already.
- `ontology_tool.py:809`: `sdk = UiPath()` inside `_ensure_graph` — this instantiates a new SDK client on every first-call. Is there a reason not to accept the SDK as a constructor parameter for testability?

### Verdict

**Approve with minor suggestions.** The architecture is sound, security model is correct, and the feature is properly gated. The duplicated sub-graph is the main thing worth discussing — it's defensible now but should have a convergence plan. The `rdflib` dependency is justified by the robustness gains over regex parsing.

milind-jain-uipath (approved):

PR #917, opened 2026-06-17
feat: caller-owned trace_context_extra_baggage; stop defaulting source=agents (v0.11.19)

RunnanJia:
## Why
The chat-model factory hardcoded `extra_baggage=["source=agents"]` when building LLM Gateway trace headers. That is incorrect for any caller that isn't agents — `source` (and `executionType`/`jobKey`) are product-owned attribution values and must be supplied by the caller, not defaulted in this shared package.

## What
- Removed the hardcoded `source=agents`.
- Added a `trace_context_extra_baggage: list[str] | None` parameter to `get_chat_model`, threaded through the new path, the legacy path, and into the OpenAI/Vertex/Bedrock legacy transports down to the per-request httpx hook (`_inject_trace_context_headers`).
- Version bump → **0.11.19**.
- Trace-context callback/forwarding tests added.

Callers (e.g. uipath-agents-python) now pass their own `source`/`executionType`/`jobKey` via this parameter.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-langchain"
-version = "0.13.18"
+version = "0.13.19"
 description = "Python SDK that enables developers to build and deploy LangGraph agents to the UiPath Cloud Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-langchain"
-version = "0.13.18"
```
ctiliescuuipath, line 3: after platform version is released icrease here also the platform version

PR #943, opened 2026-06-25
fix: categorize provider unsupported-MIME rejection as User

cotovanu-cristian:
## Summary
When the LLM client reports an unsupported attachment MIME type, categorize it as a user file error instead of an unknown runtime failure.

## What changed
- Added `uipath-llm-client>=1.16.2,<1.17.0` and bumped `uipath-langchain` to `0.14.1` so the agent can consume `UiPathLLMErrorCode`.
- Added `raise_for_llm_client_error()` in `src/uipath_langchain/agent/exceptions/llm.py` to map `UNSUPPORTED_MIME_TYPE` to `AgentRuntimeError(FILE_ERROR, USER)` while preserving provider detail.
- Updated `create_llm_node()` to handle normalized `UiPathError` separately; legacy raw exceptions still only use `as_uipath_error()` for HTTP-shaped provider errors.
- Kept multimodal file-block creation permissive: unsupported or empty MIME values still pass through to the provider, with empty MIME defaulting to `application/octet-stream`.

## Key decisions
- Recognition stays in `uipath-llm-client`; `uipath-langchain` consumes the stable `UiPathLLMErrorCode` instead of parsing provider message strings.
- Legacy raw non-HTTP exceptions are not remapped here; the unsupported-MIME path is handled only after the LLM client returns a normalized `UiPathError`.

## How it flows
```mermaid
flowchart LR
  A[model.ainvoke] --> B{raised}
  B -->|UiPathAPIError| C[raise_for_provider_http_error]
  B -->|UiPathError| D[raise_for_llm_client_error]
  D -->|UNSUPPORTED_MIME_TYPE| E[USER FILE_ERROR]
  B -->|raw Exception| F[as_uipath_error]
  F -->|UiPathAPIError| C
  F -->|other| G[re-raise original]
```

## Test plan
- [x] `uv run pytest tests/agent/multimodal/test_utils.py tests/agent/react/test_llm_node.py::TestLLMNodeProviderErrorHandling tests/agent/test_exception_helpers.py` — 48 passed, 1 warning
- [x] `uv run ruff check src/uipath_langchain/agent/exceptions/llm.py src/uipath_langchain/agent/react/llm_node.py src/uipath_langchain/agent/multimodal/invoke.py tests/agent/react/test_llm_node.py tests/agent/multimodal/test_utils.py tests/agent/test_exception_helpers.py` — all checks passed
- [x] `uv run ruff format --check src/uipath_langchain/agent/exceptions/llm.py src/uipath_langchain/agent/react/llm_node.py src/uipath_langchain/agent/multimodal/invoke.py tests/agent/react/test_llm_node.py tests/agent/multimodal/test_utils.py tests/agent/test_exception_helpers.py` — 6 files already formatted
- [x] `git diff --check` — no whitespace errors

**Breaking changes:** none

### DIFF
--- a/src/uipath_langchain/agent/multimodal/invoke.py
+++ b/src/uipath_langchain/agent/multimodal/invoke.py
@@ -61,11 +61,10 @@ async def build_file_content_blocks_for(
     if is_image(file_info.mime_type):
         return [create_image_block(base64=base64_file, mime_type=file_info.mime_type)]
 
-    mime_type = file_info.mime_type or "application/octet-stream"
     return [
         create_file_block(
             base64=base64_file,
-            mime_type=mime_type,
+            mime_type=file_info.mime_type or "application/octet-stream",
             filename=sanitize_filename(file_info.name),
         )
     ]
--- a/src/uipath_langchain/agent/react/llm_node.py
+++ b/src/uipath_langchain/agent/react/llm_node.py
@@ -11,7 +11,7 @@ from langchain_core.messages import (
 from langchain_core.tools import BaseTool
 from pydantic import BaseModel
 from uipath.agent.react import RAISE_ERROR_TOOL
-from uipath.llm_client import UiPathAPIError
+from uipath.llm_client import UiPathAPIError, UiPathError
 from uipath.llm_client.utils.exceptions import as_uipath_error
 from uipath.runtime.errors import UiPathErrorCategory
 
@@ -19,6 +19,7 @@ from uipath_langchain.chat.handlers import get_payload_handler
 
 from ..exceptions import AgentRuntimeError, AgentRuntimeErrorCode
 from ..exceptions.licensing import raise_for_provider_http_error
+from ..exceptions.llm import raise_for_llm_client_error
 from ..messages.message_utils import replace_tool_calls
 from ..tools.static_args import StaticArgsHandler
 from .constants import (
@@ -124,6 +125,9 @@ def create_llm_node(
         except UiPathAPIError as e:
             # New LLM clients surface provider HTTP errors as a normalized UiPathAPIError directly.
             raise_for_provider_http_error(e)
+        except UiPathError as e:
+            raise_for_llm_client_error(e)
+            raise
         except Exception as e:
             # Legacy in-repo clients (use_new_llm_clients=False) raise raw provider SDK exceptions.
             # Normalize via as_uipath_error and apply the same mapping when the error is HTTP-shaped; non-HTTP errors propagate.
--- a/tests/agent/react/test_llm_node.py
+++ b/tests/agent/react/test_llm_node.py
@@ -12,7 +12,8 @@ from langchain_core.messages.content import create_text_block, create_tool_call
 from langchain_core.tools import BaseTool
 from langchain_openai import AzureChatOpenAI
 from uipath.agent.react import END_EXECUTION_TOOL, RAISE_ERROR_TOOL
-from uipath.llm_client import UiPathAPIError
+from uipath.llm_client import UiPathAPIError, UiPathError, UiPathLLMErrorCode
+from uipath.runtime.errors import UiPathErrorCategory
 
 from uipath_langchain.agent.exceptions.exceptions import (
     AgentRuntimeError,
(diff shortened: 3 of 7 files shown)


src/uipath_langchain/agent/exceptions/attachments.py
```diff
@@ -0,0 +1,45 @@
...
+            raise AgentRuntimeError(
+                code=AgentRuntimeErrorCode.FILE_ERROR,
+                title="Unsupported file attachment format.",
+                detail=(
+                    "An attachment has a file type the model does not support. "
+                    "Remove the attachment or convert it to a supported format."
+                ),
+                category=UiPathErrorCategory.USER,
+            ) from e
+        cause = cause.__cause__
```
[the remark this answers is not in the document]
cotovanu-cristian: Addressed — `detail` now includes the provider message (`Provider detail: {err.detail}`, carrying the original "Unsupported MIME type: …" text) and the error chains `from exc`. The approach also changed after rebasing on the merged llm-client contract: recognition moved into `uipath-llm-client`, so the helper now matches `UiPathLLMErrorCode.UNSUPPORTED_MIME_TYPE` instead of parsing the raw message string.

tests/agent/react/test_llm_node.py
```diff
@@ -342,3 +344,44 @@ async def test_multiple_raise_error_calls_keeps_only_first(self):
...
+        Building the provider request locally fails for an attachment whose
+        file type the model does not accept. The raw error is a bare
+        ValueError, which would otherwise be categorized as UNKNOWN.
+        """
+        self.mock_model.ainvoke = AsyncMock(
+            side_effect=ValueError(
+                "Unsupported MIME type: application/octet-stream."
+                " Please refer to the provider documentation for supported formats."
+            )
+        )
```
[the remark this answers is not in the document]
cotovanu-cristian: Addressed — `tests/agent/test_exception_helpers.py::TestRaiseForUnsupportedAttachment` now covers the wrapped `__cause__` chain (`test_translates_code_in_cause_chain`) alongside the direct and no-op cases.

src/uipath_langchain/agent/multimodal/invoke.py
```diff
@@ -53,6 +55,8 @@ async def build_file_content_blocks_for(
         except ValueError as exc:
             raise ValueError(f"File '{file_info.name}': {exc}") from exc
 
+    mime_type = file_info.mime_type or "application/octet-stream"
```
vldcmp-uipath: seems that this assignment is set just for a default value

can we untangle this by either
- have this at the beginning of the method and use `mime_type` instead of `file.mime_type` in the rest of the code
- remove it altogether and have the `or "application/octet-stream"` in the return

src/uipath_langchain/agent/exceptions/attachments.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def raise_for_unsupported_attachment(exc: BaseException) -> None:
+    """Raise a USER ``FILE_ERROR`` if the error chain carries a normalized
+    ``UNSUPPORTED_MIME_TYPE``; no-op otherwise."""
+    for err in _iter_error_chain(exc):
+        if (
+            isinstance(err, UiPathError)
+            and err.error_code == UiPathLLMErrorCode.UNSUPPORTED_MIME_TYPE
+        ):
```
vldcmp-uipath: I think we no longer need an additional iter_error_chain in the way that we are using it, with the updated uipath-llm-client

src/uipath_langchain/agent/react/llm_node.py
```diff
@@ -125,9 +126,9 @@ async def llm_node(state: StateT):
             # New LLM clients surface provider HTTP errors as a normalized UiPathAPIError directly.
             raise_for_provider_http_error(e)
         except Exception as e:
-            # Legacy in-repo clients (use_new_llm_clients=False) raise raw provider SDK exceptions.
-            # Normalize via as_uipath_error and apply the same mapping when the error is HTTP-shaped; non-HTTP errors propagate.
+            # Legacy clients raise raw provider exceptions; normalize before mapping.
             uipath_error = as_uipath_error(e)
+            raise_for_unsupported_attachment(uipath_error)
```
vldcmp-uipath: `as_uipath_error` will give us an `UiPathError` which has already traversed the __ cause __ chain and built a client side error (if thats the case)

https://github.com/UiPath/uipath-llm-client-python/blob/bcfc30479ff8024d195cd83bcc6b1b1212f17c73/src/uipath/llm_client/utils/exceptions.py#L352-L383

by the time we catch the exception we should be able to have explicit handling, something like

```
       try:
            response = await llm.ainvoke(messages)
        except UiPathAPIError as e:
            # New LLM clients surface provider HTTP errors as a normalized UiPathAPIError directly.
            raise_for_provider_http_error(e)
        except UiPathError as uipath_error:
              if uipath_error.error_code == UiPathLLMErrorCode.UNSUPPORTED_MIME_TYPE:
                raise_for_unsupported_attachment(uipath_error)
              raise
        except Exception as e:
            # Legacy clients raise raw provider exceptions; normalize before mapping.
            uipath_error = as_uipath_error(e)
            if isinstance(uipath_error, UiPathAPIError):
                raise_for_provider_http_error(uipath_error)
            
            # fallback to re-raising if nothing handled
            raise
```

maybe as an enhancement in the future we can have an `UiPathClientSideError` in uipath-llm-client

src/uipath_langchain/agent/multimodal/invoke.py
cotovanu-cristian: done

vldcmp-uipath (approved):

PR #944, opened 2026-06-25
Add span for escalation memory ingestion

mjnovice:
## What

Adds a dedicated OpenTelemetry span around the **escalation memory ingestion** call (`sdk.memory.escalation_ingest_async`) in `_ingest_escalation_memory`.

This is the write-side counterpart to the lookup spans added in #865 ("Add spans for escalation memory lookup"). Previously the ingest path had no span of its own — it only set attributes (`fromMemory`, `savedToMemory`) on whatever ambient span happened to be current. Now ingestion gets its own `Save escalation memory` span so the write shows up as a first-class step in the trace / Studio UI.

## Details

- Opens a `Save escalation memory` span (`openinference.span.kind = agentMemoryWrite`) carrying `memorySpaceId`, `strategy`, `uipath.custom_instrumentation`, and `fromMemory`/`savedToMemory`.
- On failure, records the exception and sets the span status to ERROR — while still swallowing the error so a failed write never breaks the run (unchanged behavior).
- Matches the existing lookup-span pattern in the same file: local OTel import, `tracer.start_as_current_span(...)` with a `_noop_context()` fallback when tracing isn't installed.
- Bumps version `0.13.9` → `0.13.10`.

## Tests

- `test_adds_memory_write_span` — asserts the span name, kind, and attributes on a successful ingest.
- `test_sets_memory_write_span_to_error_on_ingest_failure` — asserts `savedToMemory=False`, the recorded exception, and ERROR status on a failed ingest.

All 58 tests in `tests/agent/tools/test_escalation_memory.py` pass; `ruff format`, `ruff check`, and `mypy` are clean.

### DIFF
--- a/src/uipath_langchain/agent/tools/escalation_memory.py
+++ b/src/uipath_langchain/agent/tools/escalation_memory.py
@@ -573,6 +573,7 @@ async def _ingest_escalation_memory(
     trace_id: str,
     user_id: str | None = None,
     folder_path: str | None = None,
+    memory_space_name: str | None = None,
 ) -> None:
     """Persist a resolved escalation outcome into memory."""
     normalized_user_id = _normalize_user_id(user_id)
@@ -602,6 +603,7 @@ async def _ingest_escalation_memory(
                 "type": "agentMemoryWrite",
                 "span_type": "agentMemoryWrite",
                 "uipath.custom_instrumentation": True,
+                "memorySpaceName": memory_space_name or "",
                 "memorySpaceId": memory_space_id,
                 "strategy": ESCALATION_MEMORY_STRATEGY,
             },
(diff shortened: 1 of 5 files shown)


src/uipath_langchain/agent/tools/escalation_memory.py
```diff
@@ -584,32 +583,68 @@ async def _ingest_escalation_memory(
...
+    ingest_span_ctx = (
+        tracer.start_as_current_span(
+            "Save escalation memory",
+            attributes={
+                "openinference.span.kind": "agentMemoryWrite",
+                "type": "agentMemoryWrite",
+                "span_type": "agentMemoryWrite",
+                "uipath.custom_instrumentation": True,
+                "memorySpaceId": memory_space_id,
+                "strategy": ESCALATION_MEMORY_STRATEGY,
```
smflorentino, line 608: Include memory space name too AND what was saved

mjnovice: Addressed in 46db1fb1. The "Save escalation memory" span now carries `memorySpaceName` (threaded through from the escalation tool, mirroring the lookup span) plus a `request` attribute serializing what was persisted (answer + attributes) so the Studio UI can display the saved memory item.

smflorentino (approved):

PR #959, opened 2026-07-01
fix: route AnthropicMessages apiFlavor to bedrock converse in legacy factory

andreiancuta-uipath:
- We were not handling the `AnthropicMessages` flavor until now
    - There was an `AnthropicClaude` flavor defined in the SDK, but it does not correspond to any real flavor in LLMGW and ended in an error raise anyway. I've simplified the raise
    - The new LLM client package uses a real Messages client; the legacy factory does not have such a client, so we route to the bedrock one.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-langchain"
-version = "0.13.17"
+version = "0.13.18"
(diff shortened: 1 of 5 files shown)


andreiancuta-uipath:
Does not actually work. Will fix by enabling the new LLM clients FF

PR #963, opened 2026-07-02
fix: escalation memory spans render wrong type and empty results

mjnovice:
## Problem

Customer feedback (Schneider, via #dev-agent-memory): the escalation memory span in the execution trace shows type **"Dynamic few-shot"** and nothing in the **Results** tab.

Root cause, verified against `@uipath/traceview` (2.45.0 and current `traceview-ui` main):

1. The escalation lookup emitted an "Apply escalation memory" child span typed `applyDynamicFewShot`, borrowed from the episodic memory flow. The UI labels that type "Dynamic few-shot".
2. The UI renderer for `applyDynamicFewShot` only displays episodic feedback items — it requires `response.results[].feedback.comment`. Escalation search results are shaped `results[].answer.{output, outcome}` (no `feedback` field ever), so the Results tab could never show anything.
3. Latent bug: the "Save escalation memory" ingest span used kind `agentMemoryWrite`, which is not in the UI span-type registry and renders as type **"N/A"**. The registered type is `agentMemoryStore`.

The LlmOps backend does not key any processing off these span kinds — this is purely a UI rendering contract.

## Changes

- **Lookup**: drop the `applyDynamicFewShot` child span and collapse onto the single "Find previous memories" span (`agentMemoryLookup` → renders as "Memory lookup"), now carrying `request`, `response`, `fromMemory`, `memoryItemsMatched`, and `result`. On a cache hit the cached answer also remains visible as the escalation tool span's own output.
- **Ingest**: retype `agentMemoryWrite` → `agentMemoryStore` ("Memory store"), and emit `inputs`/`outputs` attributes — exactly what the UI's default renderer displays in the Results tab — so the save span shows the persisted memory item and `savedToMemory` outcome.

## Notes

With today's traceview there is no span type that renders escalation-memory *matches* in a dedicated Results tab (`agentMemoryLookup` is deliberately excluded from that tab, same as episodic lookups). Rich rendering of the matched item needs a small `traceview-ui` follow-up (e.g. a renderer for `results[].answer` or a new `applyEscalationMemory` type). This PR fixes the wrong label and the misleading empty tab within the current contract.

## Testing

- `uv run pytest tests/agent/` — 1273 passed
- `uv run ruff check` / `ruff format --check` / `mypy` on touched files — clean

### DIFF
--- a/src/uipath_langchain/agent/tools/escalation_memory.py
+++ b/src/uipath_langchain/agent/tools/escalation_memory.py
@@ -121,7 +121,13 @@ class EscalationMemoryRetriever:
             otel_trace = None  # type: ignore[assignment]
 
         # Span attribute keys matching what the LlmOpsHttpExporter and
-        # Studio UI expect. "openinference.span.kind" sets SpanType.
+        # Studio UI expect. "openinference.span.kind" sets SpanType;
+        # "agentMemoryLookup" renders as "Memory lookup" in the trace view.
+        # Unlike episodic memory, escalation memory emits no
+        # "applyDynamicFewShot" child span: that span type's renderer only
+        # displays episodic feedback items (response.results[].feedback),
+        # which escalation results (results[].answer) never contain, so the
+        # child rendered as an empty "Dynamic few-shot" span.
         lookup_span_ctx = (
             tracer.start_as_current_span(
                 "Find previous memories",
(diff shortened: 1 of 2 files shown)


RunnanJia (approved):

saksharthakkar:
Do you need to bump up langchain version?

saksharthakkar (approved):

PR #965, opened 2026-07-02
feat: add conversational-output node for low-code conversational-agent loop

maxduu:
## Summary

Adds a dedicated graph node that runs after the regular loop for low-code conversational agent when structured-outputs are defined.

In the past, conversational agents terminated with just `uipath__agent_response_messages` once an LLM-call produces no tool-calls. Now, if an output-schema is defined and conversational-outputs are enabled, the graph routes to a new `GENERATE_CONVERSATIONAL_OUTPUT` graph node which calls the LLM to produce a structured output via a focused, forced tool call to `set_conversational_output`. 

### Why a separate tool-call rather than just give the regular agent loop the `set_conversational_output` tool?
For the regular conversational-loop, every LLM-call is with "tool_choice: auto" meaning the LLM can decide to call/not-call tool(s). An original approach attempted to give the original loop's LLM calls the `set_conversational_output` tool and prompt it to call it with its last reply in its turn. However, this causes some issues:
- GPT models oftentimes don't call the tool
- The model needs to focus on both the conversation response _and_ calling the tool, meaning possible negative effects on the response quality.

Instead, this change is purely additive - after the conversational loop (which has no context of the `set_conversational_output` tool, we make an LLM-call with a forced (tool_choice: any) call to generate the structured output. This means every model is guaranteed to call the tool and it decouples conversational quality from schema compliance — the main loop's LLM calls stay focused on the reply/tool-calls, and a second LLM call reliably fills in the structured output fields.

Note that this does have an effect on latency - however, a planned optimization (follow-up) is to ensure we emit `end_exchange` after the last regular loop's LLM-call rather than after the entire graph.

## End-to-end example:
In addition to the original `uipath_agent_response_messages` output which is always implicit, `outputSchema` fields will be respected and the agent can output things like:
`agent.json`:
```
"messages": [
        {
            "role": "system",
            "content": "You are an agentic assistant that searches the web and can analyze file attachments. If the user asks questions related to service desk tickets, route to the service-desk agent. If the user mentions that they want to speak to a human, route to a human.",
          ...
    ],
    "inputSchema": {
        "type": "object",
        "properties": {}
    },
    "outputSchema": {
        "type": "object",
        "properties": {
            "should_handoff": {
                "type": "boolean",
                "description": "Whether the conversation should be handed off to a different agent."
            },
            "handoff_target": {
                "type": "string",
                "description": "If should_handoff is true, this field indicates the target to handoff to. Either \"human\" or \"service_desk_agent\"."
            }
        }
    },
```
Output:
```
● END
output
├── should_handoff: True
├── handoff_target: human
└── uipath__agent_response_messages (list, 1 items)
    └── #0 (dict)
        ├── role: assistant
        ├── contentParts (list, 1 items)
        │   └── #0 (dict)
        │       ├── mimeType: text/markdown
        │       ├── data (dict)
        │       │   └── inline: I will connect you to a human representative. Please wait while I transfer your request.
        │       └── citations (list, 0 items)
        └── toolCalls (list, 0 items) 
```

## Changes

**Graph structure** ([agent/react/](src/uipath_langchain/agent/react/))
- New `AgentGraphNode.GENERATE_CONVERSATIONAL_OUTPUT` enum value.
- New `conversational_output_node.py` — invokes a focused LLM call with only `set_conversational_output` bound (`tool_choice="any"`, `disable_streaming=True`, `TAG_NOSTREAM` tag). Reuses `state.messages` for full agent context and appends the framework instruction as a HumanMessage that never persists to state.
- `agent.py` — conditionally inserts the new node between AGENT and TERMINATE when its a conversational agent and `has_custom_conversational_output_fields(output_schema)`.
- `router_conversational.py` — routes AGENT-without-tool-calls to the new node (or straight to TERMINATE if the flag is off).

**Config**
- New `AgentGraphConfig.conversational_outputs_enabled: bool = False` — the top-level feature flag. Defaults to False; existing callers unchanged.

**Terminate node**
- `_handle_end_conversational` now best-effort extracts the tool call's args from `state.messages[-1]`. If the tool call is absent, custom fields stay empty and Pydantic surfaces a clear per-field error at schema validation.
- The new node's AIMessage is sliced off before conversion so it doesn't leak into the response payload.

**Utilities**
- New `has_custom_conversational_output_fields` + `build_conversational_output_args_schema` helpers in `utils.py` (strips `uipath__agent_response_messages` from the LLM-fillable args schema).
- New `create_set_conversational_output_tool` factory in `agent/react/tools/tools.py`.
- New shared `config_without_streaming` helper in `agent/tools/utils.py` — refactored out of `analyze_files_tool.py` since it's now used in two places.
- New `UIPATH_CONVERSATIONAL_AGENT_RESPONSE_MESSAGES_FIELD` constant.

**Tests**: full topology + router + terminate + utils + node coverage added.

## Related PRs

Part of a coordinated four-repo change. Each PR is independently reviewable, but they land together:

- **uipath-runtime-python** https://github.com/UiPath/uipath-runtime-python/pull/142 — surfaces the `conversationalService.enableOutputs` FPS flag on `UiPathRuntimeContext`.
- **uipath-python** https://github.com/UiPath/uipath-python/pull/1785 — adds the `set_conversational_output` flow-control tool + generate-output prompt primitives.
- **uipath-langchain-python** https://github.com/UiPath/uipath-langchain-python/pull/965 — implements the `GENERATE_CONVERSATIONAL_OUTPUT` graph node, router path, and terminate extraction (the main change).
- **uipath-agents-python** https://github.com/UiPath/uipath-agents-python/pull/588 — Gates custom outputs on the FPS flag + agent-version.

## Test plan

- [x] \`uv run pytest tests/agent tests/runtime tests/cli\` (1567 tests pass)
- [x] \`just lint\` clean
- [x] Manual end-to-end against \`convo-agent-web-search-file-attachments\`


src/uipath_langchain/agent/react/terminate_node.py
```diff
@@ -67,8 +80,34 @@ def _handle_end_conversational(
...
+        dict(set_output_call["args"]) if set_output_call is not None else {}
+    )
+
     initial_count = state.inner_state.initial_message_count
-    new_messages = state.messages[initial_count:]
+    # Drop the GENERATE_CONVERSATIONAL_OUTPUT AIMessage from the converted
+    # response payload — it carries no user-visible content, only the
+    # framework-internal set_conversational_output tool call.
+    new_messages = (
+        state.messages[initial_count:-1]
```
andrewwan-uipath: If the extraction call comes back with no `set_conversational_output` tool call, will the empty extraction message leak into `uipath__agent_response_messages` as a blank assistant turn?

maxduu: Good call - I did a bit of refactoring here https://github.com/UiPath/uipath-langchain-python/pull/965/commits/9302f95ce3be471245f6f4b367d9f506955ddec8 to be more explicit. It now passes in `with_conversational_output_node` to the terminate-node, so it only parses/extracts the last-message when its guaranteed to have the `conversational_output_node` in the graph; it also throws errors when its expected to have a last AI-message with the set_conversational_output tool but there isn't one.

Could you take another look?

src/uipath_langchain/agent/react/conversational_output_node.py
```diff
@@ -0,0 +1,101 @@
...
+                detail=(
+                    "The language model returned an unexpected response type. "
+                    "If you are using a BYOM configuration, verify your model deployment.",
+                ),
+                category=UiPathErrorCategory.SYSTEM,
+            )
+
+        payload_handler.check_stop_reason(response)
+
+        return {"messages": [response]}
```
andrewwan-uipath, line 99: Do we have a ToolMessage result added for this? Since it's returned into graph state, will it stay in the conversation history and get replayed on the next turn as a tool call with no matching result? Would that be an issue?

maxduu: So this works the same way as the `end_execution` tool, where the LLM adds the `AIMessage` with the tool-call inputs. The tool itself is never executed since we immediately route to the TERMINATE node, so there's no `ToolMessage` result.

To ensure this `AIMessage` is not added to chat-history, we use `config_without_streaming` [here](https://github.com/UiPath/uipath-langchain-python/pull/965/changes#diff-fa8bd0e415cfd606382dbf28f239354cd0483695040fc84c2cf7c3c9ed50a76fR68) - that makes it so it doesn't stream any events to CAS and thus won't be added to the conversation-history. This is the same pattern used for the analyze-attachments tool which also makes an LLM-call that isn't streamed to CAS.

Finally, we ensure the `AIMessage` isn't part of the `uipath__agent_response_messages` by parsing it out [here](https://github.com/UiPath/uipath-langchain-python/pull/965/changes#diff-f355d4521b00a7cbb121a644ef59c422253af4875a8bdac740766ba17e45755eR117) with `new_conversation_messages = state.messages[initial_count:-1]`.

So in summary - no `ToolMessage` since the tool is never executed (route directly to TERMINATE node) and then the AIMessage is never streamed to CAS, and additionally not part of the output because we extract it out.

andrewwan-uipath (approved):

PR #977, opened 2026-07-08
feat: add skills feature to advanced agent with workspace to load skills

mariadhakalUipath:
Title: feat: add skills support to advanced agents with sandboxed workspace tools

Summary

Adds skills feature for advanced agents. Agent can act on it — run uip commands, review what it ran, and verify what it scaffolded.

### DIFF
--- a/src/uipath_langchain/agent/advanced/agent.py
+++ b/src/uipath_langchain/agent/advanced/agent.py
@@ -24,6 +24,7 @@ from .utils import (
     MEMORY_INDEX_VIRTUAL_PATH,
     create_state_with_input,
     resolve_input_attachments,
+    resolve_skill_sources,
 )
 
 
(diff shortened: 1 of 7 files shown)


src/uipath_langchain/agent/advanced/agent.py
```diff
@@ -50,6 +57,7 @@ def create_advanced_agent(
         backend=backend,
         response_format=response_format,
         memory=list(memory) or None,
+        skills=list(skills) if skills is not None else None,
```
gauravsingh-rajput: because of the default None in the function call this could be simply `skills = list(skills)`

mariadhakalUipath: Changed skills = list(skills) or None instead to avoid triggering deepagents skillsmidleware when the skills list is passed empty.

PR #983, opened 2026-07-13
fix: tolerate malformed tool output schemas at agent startup

UiPathPetruPopa:
Agents that ran on the legacy Temporal/.NET runtime crash on URT at startup with
`AGENT_STARTUP.INVALID_TOOL_CONFIG` when a tool's **output** schema has a dangling
`$ref` (e.g. a .NET `decimal?` → `Nullableofdecimal` with no `$defs`). URT eagerly
converts every tool schema to Pydantic and throws; the old runtime never resolved
`$ref`s. Output schemas are design-time only (guardrails + eval sims), never used
during execution, so the crash is gratuitous.

- Added `create_output_model()` in `jsonschema_pydantic_converter.py` — falls back
  to an empty model (with a warning) instead of raising on an unparseable schema.
- Applied it to the 5 output-schema tool factories: `process`, `escalation`,
  `integration`, `analyze_files`, `deeprag`.
- Left fatal (intentional): all **input** schemas and the agent-level output schema
  in `react/utils.py` (used at runtime for terminate-node validation).

### DIFF
--- a/src/uipath_langchain/agent/react/jsonschema_pydantic_converter.py
+++ b/src/uipath_langchain/agent/react/jsonschema_pydantic_converter.py
@@ -1,4 +1,5 @@
 import inspect
+import logging
 import sys
 from types import ModuleType
 from typing import Any, Type, cast
@@ -7,6 +8,12 @@ from jsonschema_pydantic_converter import transform_with_modules
 from pydantic import BaseModel, PydanticUndefinedAnnotation
 
 from uipath_langchain.agent.exceptions import AgentStartupError, AgentStartupErrorCode
+
+logger = logging.getLogger(__name__)
+
+# Empty, always-parseable output model used as a non-fatal fallback
+# (see create_output_model).
+_EMPTY_OUTPUT_SCHEMA: dict[str, Any] = {"type": "object", "properties": {}}
 
 # Shared pseudo-module for all dynamically created types
 # This allows get_type_hints() to resolve forward references
@@ -60,3 +67,27 @@ def create_model(
     model.__module__ = _DYNAMIC_MODULE_NAME
 
     return model
+
+
+def create_output_model(
+    schema: dict[str, Any],
+    tool_name: str,
+) -> Type[BaseModel]:
+    """Convert a tool's OUTPUT JSON schema to a Pydantic model, non-fatally.
+
+    Unlike input schemas, a tool's output schema is used only at design time
+    (guardrail definitions and eval simulations) and is never used to validate
+    output during execution.
+
+    Returns:
+        The converted model, or an empty model if the schema is unparseable.
+    """
+    try:
+        return create_model(schema)
+    except AgentStartupError as e:
+        logger.warning(
+            "Tool %r has an unparseable output schema; ignoring it (non-blocking): %s",
+            tool_name,
+            e.error_info.detail,
+        )
+        return create_model(_EMPTY_OUTPUT_SCHEMA)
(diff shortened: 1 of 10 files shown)


tests/agent/tools/test_tool_factory.py
```diff
@@ -405,3 +465,37 @@ async def test_function_resource_routes_through_process_tool_path(
...
+        ],
+    )
+    async def test_malformed_output_schema_is_non_blocking(
+        self, resource_fixture, mock_uipath_sdk, request
+    ):
+        """Each of the 5 tool factories that convert an output schema must tolerate
+        a malformed one instead of failing agent startup.
+
+        Output schemas are design-time only (guardrails + eval simulation) and are
+        never used during execution, so a dangling ``$ref`` (e.g. a .NET
```
andreitava-uip: nit: technically not correct. The output model/schema is being used at runtime, but not for the core tool calling flow.
In particular it's used for:
- job attachment discovery
- applying guardrails
- simulations

src/uipath_langchain/agent/react/jsonschema_pydantic_converter.py
```diff
@@ -60,3 +67,27 @@ def create_model(
...
+    """
+    try:
+        return create_model(schema)
+    except AgentStartupError as e:
+        logger.warning(
+            "Tool %r has an unparseable output schema; ignoring it (non-blocking): %s",
+            tool_name,
+            e.error_info.detail,
+        )
+        return create_model(_EMPTY_OUTPUT_SCHEMA)
```
andreitava-uip: This is actually trickier than first expected. Output model is used in places like detecting job attachments, and replacing malformed schemas with empty can lead to silent runtime failures that would be incredibly difficult to debug.

andreitava-uip (changes requested):
We have to think of a different fallback, Ideally only remove the malformed parts and keep the rest of the fields?

src/uipath_langchain/agent/react/jsonschema_pydantic_converter.py
```diff
@@ -60,3 +74,102 @@ def create_model(
...
+        # Last-resort net for a non-$ref failure we didn't neutralize. Intentionally
+        # narrow (AgentStartupError only): other errors are unexpected and should
+        # surface rather than be silently swallowed.
+        logger.warning(
+            "Tool %r output schema still unparseable after neutralizing dangling "
+            "refs; falling back to an empty model (non-blocking): %s",
+            tool_name,
+            e.error_info.detail,
+        )
+        return create_model(_EMPTY_OUTPUT_SCHEMA)
```
andreitava-uip: I'd remove this last resort fallback. The dangling refs fix is good because it (mostly) preserves functionality such as file attachments and guardrails.

For more fatal issues, I'd rather throw up-front than run into obscure silent bugs at runtime.

Most malformed issues we discovered were such dangling refs. Until further notice I wouldn't add extra fallbacks.

andreitava-uip (changes requested):

src/uipath_langchain/agent/react/jsonschema_pydantic_converter.py
UiPathPetruPopa: Removed last resort fallback.

andreitava-uip (approved):

src/uipath_langchain/agent/react/jsonschema_pydantic_converter.py
dianagrecu-uipath, line 56: nit: Is this message applicable here since this case is treated in the create_output_model method?

UiPathPetruPopa: This message lives in create_model, the shared converter used by all the other paths (input schemas, the agent-level output schema, static args, etc.). A dangling $ref in any of those still raises it, and the message accurately describes the fix — so I'd keep it as is. My change in create_output_model only neutralizes dangling refs for a tool's output schema.

PR #989, opened 2026-07-14
feat(skills): add skills support to advanced agents with sandboxed cli tool

mariadhakalUipath:
Title: Uipath Skills with Cli tool for advanced agents. 

Summary

Adds skills feature for advanced agents. Agent can act on it — run uip commands, review what it ran, and verify what it scaffolded.

⚠️ Note for coded agents

The feature flag exists for low-code graph builder reached only via `agent.json` →
`AgentsRuntimeFactory` → `_build_advanced_agent`. 

**Coded agents resolve to a different runtime factory (`langgraph.json`) and never evaluate that flag.** A coded author who writes:
`from uipath_langchain.agent.tools.internal_tools import create_uipath_cli_tool`

gets full `uip` execution with no kill switch at this layer.

If you are adding this tool to a coded agent and need destructive commands restricted, wrap it with your own validation before handing it to the agent. Do not rely on a tool-level `@guardrail`, which does not fire for this tool.
For example, delegate to the tool from your own `StructuredTool` that rejects the commands you care
about, and apply platform guardrails at the agent/graph level rather than the tool level.

### DIFF
--- a/src/uipath_langchain/agent/advanced/agent.py
+++ b/src/uipath_langchain/agent/advanced/agent.py
@@ -87,12 +87,16 @@ def create_advanced_agent(
     response_format: ResponseFormat[Any] | None = None,
     memory: Sequence[str] = (),
     middleware: Sequence[AgentMiddleware[Any, Any]] = (),
+    skills: Sequence[str] | None = None,
 ) -> CompiledStateGraph[Any, Any, Any, Any]:
     """Create a deepagents agent with planning, filesystem, and sub-agent tools.
 
     ``memory`` is a list of file paths loaded via deepagents' ``MemoryMiddleware``:
     each is read from ``backend`` and injected into the system prompt every turn,
     and the model maintains them with ``edit_file``. Empty disables the middleware.
+
+    ``skills`` is a list of skill source paths for deepagents' ``SkillsMiddleware``;
+    ``None`` or empty disables it (mirroring ``_create_deep_agent``'s contract).
     """
     return _create_deep_agent(
         model=model,
@@ -103,6 +107,7 @@ def create_advanced_agent(
         response_format=response_format,
         memory=list(memory) or None,
         middleware=list(middleware),
+        skills=list(skills) if skills else None,
     )
 
 
@@ -115,6 +120,7 @@ def create_advanced_agent_graph(
     input_schema: type[BaseModel] | None,
     output_schema: type[BaseModel],
     build_user_message: Callable[[dict[str, Any]], str],
+    skills: Sequence[str] | None = None,
 ) -> StateGraph[Any, Any, Any, Any]:
     """Wrap the advanced agent in a parent graph that maps typed I/O to/from messages.
 
@@ -153,6 +159,7 @@ def create_advanced_agent_graph(
             if runtime_system_prompt_key is not None
             else []
         ),
+        skills=skills,
     )
 
     wrapper_state = create_state_with_input(input_schema)
@@ -215,6 +222,7 @@ def create_conversational_advanced_agent_graph(
     tools: Sequence[BaseTool],
     system_prompt: str,
     backend: BackendProtocol | BackendFactory | None,
+    skills: Sequence[str] | None = None,
 ) -> StateGraph[Any, Any, Any, Any]:
     """Wrap the advanced agent in a parent graph that speaks the conversational contract.
 
@@ -237,6 +245,7 @@ def create_conversational_advanced_agent_graph(
         system_prompt=system_prompt,
         backend=backend,
         memory=memory_sources,
+        skills=skills,
     )
 
     class ConversationalAdvancedAgentOutput(BaseModel):
--- a/src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
+++ b/src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
@@ -1,0 +1,259 @@
+"""Sandboxed ``uip`` CLI runner injected into advanced agents.
+
+Unlike the resource-selected internal tools built by ``create_internal_tool``, this
+tool is injected programmatically by the agent graph builder when UiPath skills are
+active — the agent invokes it to run the ``uip`` commands a skill prescribes. It runs
+exactly one ``uip`` invocation per call.
+
+"""
+
+import asyncio
+import logging
+import os
+import shlex
+import shutil
+from pathlib import Path
+from typing import Any
+
+from langchain_core.tools import StructuredTool
+from pydantic import BaseModel, Field
+from uipath.eval.mocks import mockable
+from uipath.runtime import Workspace
+
+from uipath_langchain.agent.tools.structured_tool_with_argument_properties import (
+    StructuredToolWithArgumentProperties,
+)
+
+logger = logging.getLogger("uipath")
+
+_TOOL_NAME = "uipath_cli"
+
+_TOOL_DESCRIPTION = (
+    "Run a single UiPath `uip` command in the agent workspace; returns "
+    "exit_code/stdout/stderr. Only the uip/uipath binaries, one command per call; "
+    "shell chaining is refused. A negative exit_code means the command was refused "
+    "before it ran and stderr explains why. Use `subdir` to target a scaffolded "
+    "project folder."
+)
+
+_ALLOWED_BINARIES: tuple[str, ...] = ("uip", "uipath")
+
+_REJECTED_EXIT_CODE = -1000
+
+_COMMAND_TIMEOUT_SECONDS = 600
+
+_SHELL_OPERATORS: frozenset[str] = frozenset({"&&", "||", ";", "|"})
+
+_COMMAND_EXAMPLES: tuple[str, ...] = (
+    "solution pack",
+    "solution publish",
+    "codedagent init",
+)
+
+
+def _parse_uip_command(command: str) -> list[str]:
+    """Validate a single ``uip`` command and return its argv (binary excluded).
+
+    Args:
+        command: The command to run, e.g. ``"pack"`` or ``"solution publish"``.
+            An explicit ``uip``/``uipath`` prefix is tolerated and stripped.
+
+    Returns:
+        The argument tokens to pass after the resolved binary.
+
+    Raises:
+        ValueError: If the command is empty or chains multiple commands via a shell
+            operator.
+    """
+    try:
+        lexer = shlex.shlex(command, posix=True)
+        lexer.whitespace_split = True
+        lexer.commenters = ""
+        lexer.escape = ""
+        tokens = list(lexer)
+    except ValueError as exc:
+        raise ValueError(f"Could not parse command: {exc}") from exc
+
+    if not tokens:
+        raise ValueError("No command provided.")
+
+    if tokens[0] in _ALLOWED_BINARIES:
+        tokens = tokens[1:]
+    if not tokens:
+        raise ValueError("No `uip` subcommand provided.")
+
+    for token in tokens:
+        if token in _SHELL_OPERATORS:
+            raise ValueError(
+                f"Shell operator '{token}' is not allowed; run one command at a time."
+            )
+
+    return tokens
+
+
+class UiPathCliInput(BaseModel):
+    """Input schema for the ``uipath_cli`` tool."""
+
+    command: str = Field(
+        description=(
+            "A single `uip` command without the binary prefix, e.g. `solution pack`, "
+            "`solution publish`. One command only; no shell operators."
+        ),
+        examples=list(_COMMAND_EXAMPLES),
+    )
+    subdir: str = Field(
+        default="",
+        description="Workspace-relative directory to run in; defaults to the root.",
+    )
+
+
+class UiPathCliOutput(BaseModel):
+    """Output schema for the ``uipath_cli`` tool."""
+
+    command: str = Field(description="The argv actually executed (or attempted).")
+    exit_code: int = Field(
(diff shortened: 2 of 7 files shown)


src/uipath_langchain/agent/advanced/agent.py
gauravsingh-rajput, line 1: `_create_deep_agent` treats `skills=None` as the default/disabled state, can we keep that same shape through our wrapper APIs? Right now `create_advanced_agent_graph` accepts None, converts it to (), and then `create_advanced_agent` converts it back to None. We can simplify by defaulting skills to None in all three functions, passing `skills=skills` through the graph wrappers, and only converting non-empty sequences to a list at the `_create_deep_agent` boundary.

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
```diff
@@ -0,0 +1,221 @@
...
+_ALLOWED_BINARIES: tuple[str, ...] = ("uip", "uipath")
+
+_REJECTED_EXIT_CODE = -1
+
+_COMMAND_TIMEOUT_SECONDS = 600
+
+_SHELL_OPERATORS: frozenset[str] = frozenset({"&&", "||", ";", "|"})
+
+# Adding this list for the first iteration of the tool, but we may not need it in the future.
+_DESTRUCTIVE_SUBCOMMANDS: frozenset[str] = frozenset(
```
gauravsingh-rajput: i think an allowlist if safer, lets allow just agent commands for now, we can add stuff later on.

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
```diff
@@ -0,0 +1,221 @@
...
+        return UiPathCliOutput(
+            command=echoed,
+            exit_code=proc.returncode
+            if proc.returncode is not None
+            else _REJECTED_EXIT_CODE,
+            stdout=stdout.decode(errors="replace"),
+            stderr=stderr.decode(errors="replace"),
+        ).model_dump()
+
+    return StructuredTool(
```
gauravsingh-rajput: so `UiPathCliOutput` is built and immediately .model_dump()-ed, and the tool is a plain `StructuredTool`. That means the output model — and its field descriptions (e.g. "a negative value means the command was rejected") — never reach the runtime/tracing/UI. Every other internal tool (analyze_files, batch_transform, deeprag) advertises its output via `StructuredToolWithArgumentProperties(output_type=…, metadata={... "output_schema": ...})`. Matching that here makes the output schema visible and keeps this tool consistent with its siblings. check [this](https://github.com/UiPath/uipath-langchain-python/blob/6f3a70513517c6349d9cbf070d3e833747d556ac/src/uipath_langchain/agent/tools/internal_tools/deeprag_tool.py#L159)

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
```diff
@@ -0,0 +1,221 @@
...
+    excluding destructive subcommands.
+
+    Args:
+        workspace: The agent's workspace; commands run within its directory.
+
+    Returns:
+        A ``StructuredTool`` named ``uipath_cli``.
+    """
+
+    async def run_uipath_command(command: str, subdir: str = "") -> dict[str, Any]:
```
gauravsingh-rajput, line 195: analyze_files, batch_transform, and deeprag all wrap their real invocation in @mockable so eval runs substitute a deterministic response instead of hitting the live operation. Without @mockable, any eval set that exercises an agent using this tool has no way to stub it. check [this](https://github.com/UiPath/uipath-langchain-python/blob/6f3a70513517c6349d9cbf070d3e833747d556ac/src/uipath_langchain/agent/tools/internal_tools/deeprag_tool.py#L109)

gauravsingh-rajput (approved):

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
```diff
@@ -0,0 +1,273 @@
...
+# signal-terminated processes surface as -1..-64, so a value well below -255
+# unambiguously marks a command rejected before it ever ran.
+_REJECTED_EXIT_CODE = -1000
+
+_COMMAND_TIMEOUT_SECONDS = 600
+
+_SHELL_OPERATORS: frozenset[str] = frozenset({"&&", "||", ";", "|"})
+
+# Adding this list for the first iteration of the tool, but we may not need it in the future.
+_DESTRUCTIVE_SUBCOMMANDS: frozenset[str] = frozenset(
```
radu-mocanu: the list of disallowed commands should sit next to the CLI so they can be versioned together. what happens if a new CLI binary is available on the serverless image (containing a destructive command that is not blackilsted here; like remove tenant or similar) but the coded agent is pinned to this SDK version?

I suggest we add support on the CLI side, a new global flag like —safe or —disallow-destructive-command and block the calls at CLI level. our harness should then always append this flag to the commands it invokes

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
```diff
@@ -0,0 +1,273 @@
...
+
+class UiPathCliInput(BaseModel):
+    """Input schema for the ``uipath_cli`` tool."""
+
+    command: str = Field(
+        description=(
+            "A single `uip` command without the binary prefix, e.g. `pack`, "
+            "`solution publish`. One command only; no shell operators."
+        ),
+        examples=["pack", "solution publish", "init my-agent"],
```
radu-mocanu: let s include valid commands in the examples here. 
eg init my-agent -> codedagent init

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
```diff
@@ -0,0 +1,242 @@
...
+
+class UiPathCliInput(BaseModel):
+    """Input schema for the ``uipath_cli`` tool."""
+
+    command: str = Field(
+        description=(
+            "A single `uip` command without the binary prefix, e.g. `pack`, "
+            "`solution publish`. One command only; no shell operators."
+        ),
+        examples=["pack", "solution publish", "codeagent init"],
```
radu-mocanu: codeagent init is not a valid command; the cli returns unknown command 'codeagent'. use codedagent init here, and add a schema assertion so the example stays aligned with the actual cli.

src/uipath_langchain/agent/advanced/agent.py
```diff
@@ -87,12 +87,16 @@ def create_advanced_agent(
     response_format: ResponseFormat[Any] | None = None,
     memory: Sequence[str] = (),
     middleware: Sequence[AgentMiddleware[Any, Any]] = (),
+    skills: Sequence[str] | None = None,
```
radu-mocanu, line 90: this is a logic-changing public change, but the pr does not bump the package version. main is already at 0.16.2 while this branch still has 0.16.1. rebase and bump to the next version in pyproject.toml before merge.

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
```diff
@@ -0,0 +1,242 @@
...
+
+            run_dir = _resolve_run_dir_within_workspace(
+                workspace.path.resolve(), subdir
+            )
+            if run_dir is None:
+                return _rejected(
+                    echoed, f"Invalid subdir '{subdir}': escapes the workspace."
+                )
+
+            return await _run_uip_subprocess(binary, args, run_dir, echoed)
```
radu-mocanu, line 242: this still executes the cli arguments directly, without the cli owning and enforcing a safe mode. the parser currently accepts commands such as solution delete and admin tenants delete, so resolving the previous thread did not address the version-skew problem. let s add the safety flag in the cli and always append it here before shipping this tool.

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+    except OSError as exc:
+        return _rejected(echoed, str(exc))
+
+    try:
+        stdout, stderr = await asyncio.wait_for(
+            proc.communicate(), timeout=_COMMAND_TIMEOUT_SECONDS
+        )
+    except asyncio.TimeoutError:
+        proc.kill()
```
radu-mocanu: proc.kill() can raise ProcessLookupError when the child exits in the timeout and kill race. i reproduced that path and the tool crashes instead of returning the recoverable timeout payload. catch that race and still await proc.wait().

src/uipath_langchain/agent/advanced/agent.py
```diff
@@ -103,6 +107,7 @@ def create_advanced_agent(
         response_format=response_format,
         memory=list(memory) or None,
         middleware=list(middleware),
+        skills=list(skills) if skills else None,
```
radu-mocanu, line 110: this tool is still not wired into the advanced agent. with skills enabled, we only forward the skill paths and leave tools unchanged, so the agent has no way to call uipath_cli. let s add it here for the filesystem-backed skills path and cover the final tool list in a test.

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
mariadhakalUipath: Thank you for correcting me. I have changed the example command and also added the schema assertion for the example commands.

src/uipath_langchain/agent/advanced/agent.py
mariadhakalUipath: Okay will do this before I merge.

src/uipath_langchain/agent/advanced/agent.py
mariadhakalUipath: We have this tool behind feature flag and we wired it when creating advanced agent graph for low-code agents if skills are enabled instead of wiring it to advanced agent. The scope for now is to attach the tool for low-code agents.

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
mariadhakalUipath: Fix: 
```
except asyncio.TimeoutError:
    try:
        proc.kill()
    except ProcessLookupError:
        pass
    await proc.wait()

```
wait_for timing out only means the child hadn't finished when the timer fired it can
exit before kill() runs, and the transport teardown that nulls _proc makes kill()
raise. await proc.wait() stays outside the try so it runs on both paths (safe after
teardown: _wait() returns the cached returncode immediately). Catching only
ProcessLookupError so a real PermissionError still surfaces.

Added Regression test: test_run_times_out_when_child_exits_during_kill to verify it fails
with an uncaught ProcessLookupError against the unpatched code.

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
mariadhakalUipath: This is really good point we were thinking about this, the reason I don't have it implemented is:
1. We are using the user's token to run these commands so it is already tenant and folder scoped, agent can only run the commands that user is allowed.
2. This tool is used by both conversational and autonomous low-code agent, conversational agent uses the interactive mode to ask user to run the command and only the autonomous low-code agent is running the command by itself so I added prompt-based guardrail to end the execution if there is any destructive command.
3. It is only attached to agent if the skills are enabled so it will only follow the instructions there and if the user is using this tool in coded agent then it becomes their responsibility to filter the commands.
Please guide me if I am missing anything that I should have considered for this, and I can work through it.

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
```diff
@@ -0,0 +1,251 @@
...
+            stderr=asyncio.subprocess.PIPE,
+        )
+    except OSError as exc:
+        return _rejected(echoed, str(exc))
+
+    try:
+        stdout, stderr = await asyncio.wait_for(
+            proc.communicate(), timeout=_COMMAND_TIMEOUT_SECONDS
+        )
+    except asyncio.TimeoutError:
```
radu-mocanu, line 168: Handle task cancellation here as well. Cancelling `proc.communicate()` only cancels the coroutine that drains the pipes and waits for completion; it does not terminate the separate OS process. I reproduced this with a real child process: after cancelling `_run_uip_subprocess`, `proc.returncode` remained `None` and the child kept running.

Add this cancellation branch alongside the timeout branch:

```python
    except asyncio.CancelledError:
        try:
            proc.kill()
        except ProcessLookupError:
            pass
        await asyncio.shield(proc.wait())
        raise
```

This kills and reaps the child, while the bare `raise` preserves the original cancellation. Add a regression test that cancels the task and asserts both `kill()` and `wait()` are called. This covers graceful asyncio cancellation only; a hard kill of the Python parent bypasses all cleanup and must be handled by the process supervisor.

radu-mocanu: FYI this protection is for graceful stopping only. serverless force kill will most likely not stop the underlying CLI command

radu-mocanu (approved):

src/uipath_langchain/agent/tools/internal_tools/uipath_cli_tool.py
mariadhakalUipath: Added the timeout branch.

PR #997, opened 2026-07-20
feat(chat): surface missing-model discovery errors as AgentStartupError

tudormatei1:
## What & why

When a model name isn't returned by the discovery API, `get_model_info` raises `ModelNotFoundError`. Until now `get_chat_model` let it propagate raw, so the runtime showed an opaque exception instead of an actionable message.

This PR catches `ModelNotFoundError` and re-raises it as a structured `AgentStartupError` (**code** `LLM_INVALID_MODEL`, **category** `DEPLOYMENT`). If a `byo_connection_id` is set, the message points to the connection's bring-your-own-model config; otherwise it asks you to check the model name and tenant enablement. The original error is preserved as `__cause__`.

## Changes

- `chat/chat_model_factory.py`: maps `ModelNotFoundError` to `AgentStartupError`. `AgentStartupError` / `UiPathErrorCategory` imported lazily inside the except block to avoid a `chat -> agent` import cycle.
- `pyproject.toml`: `uipath-llm-client` / `uipath-langchain-client` floors to `>=1.17.1,<1.18.0`; version to `0.14.11`.
- `uv.lock`: regenerated (resolves 1.17.1).
- Tests: `TestGetChatModelModelNotFound` covers the mapping, BYO variant, and `__cause__` chain.

## Depends on

`uipath-llm-client` **1.17.1** (adds `ModelNotFoundError` + `UiPathLLMErrorCode.MODEL_NOT_FOUND`). Already published.

### DIFF
--- a/src/uipath_langchain/chat/chat_model_factory.py
+++ b/src/uipath_langchain/chat/chat_model_factory.py
@@ -14,6 +14,7 @@ from typing import Any, Final
 
 from langchain_core.callbacks import BaseCallbackHandler, Callbacks
 from langchain_core.language_models import BaseChatModel
+from uipath.llm_client.utils.exceptions import ModelNotFoundError
 from uipath.llm_client.utils.headers import (
     get_dynamic_request_headers,
     set_dynamic_request_headers,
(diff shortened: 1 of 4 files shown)


src/uipath_langchain/chat/chat_model_factory.py
```diff
@@ -125,40 +126,67 @@ def get_chat_model(
...
-        vendor_type=vendor_type,
-        api_flavor=api_flavor,
-        custom_class=custom_class,
-        agenthub_config=agenthub_config,
-        **optional_kwargs,
-        **kwargs,
-    )
+        from uipath_langchain.agent.exceptions import (
+            AgentStartupError,
+            AgentStartupErrorCode,
```
vldcmp-uipath: Why exactly do we need to lazy-import the exceptions here?

> imported lazily inside the except block to avoid a chat -> agent import cycle

why does this cycle happen? need more context for this to check if its an issue

vldcmp-uipath (changes requested):

src/uipath_langchain/chat/chat_model_factory.py
tudormatei1: if i move it to the top as global import, the test_no_circular_imports.py still passes, so you're right its not needed per se. i ll move it to a global import in the next commit where i also bump version

vldcmp-uipath (approved):

PR #1007, opened 2026-07-24
test(model-onboarding): simple, tool-call, and file payloads per path

denispetre:
## What

The model-onboarding testcase now exercises **three capability payloads for every code path**, not just file processing:

| payload | what it does |
|---|---|
| `simple` | plain `ainvoke`; asserts a non-empty completion |
| `tools` | **full tool-calling round trip** — bind `get_weather`, force the call, execute the tool, feed the `ToolMessage` back, assert the final answer uses the result |
| `files/<name>` | one cell per selected file attachment (unchanged) |

So `azure_responses` (for example) now produces `simple`, `tools`, `files/image`, `files/pdf` cells, all rolled into the single `success` boolean.

## Notable

- `files` may now be empty without losing coverage — `simple` + `tools` still run.
- The `tools` payload is a **full round trip** (executes the tool and checks the final answer used the result), per request — stricter than a request-only check.
- Backward compatible: `input.json` schema is unchanged; existing specs just gain the two new cell types.

## Verification

Graph compiles and `input.json` validates against the updated schema. Round-trip logic verified with fake models covering happy path, no-tool-call, and ignored-result cases (all assert correctly). Live LLM runs happen in CI / on-demand dispatch (needs tenant auth).


denispetre:
**Added a fourth payload: `is_tools` — IS activity tools per API flavor** (commit e011df8)

Per flavor, the model under test binds an Integration Service activity tool, is forced to call it, the activity executes through a real IS connection, and the result feeds a final answer (full round trip). Cells: `is_tools/<flavor>`.

| flavor | connector / activity | body style |
|---|---|---|
| `azure_openai` | microsoft-azureopenai / generateChatCompletionConsolidated | JSON + `modelId` query |
| `openai` | openai-openai / v2::chat::completion | multipart (`body` JSON part) |
| `openai_v1` | openai-openaiv1compliant / chatCompletion | JSON |
| `bedrock_converse` | aws-bedrock / completion::converse | multipart + `modelName` query |
| `vertex` | google-vertex / textCompletionUsingGemini | multipart + `modelName` path |
| `anthropic` | anthropic-claude / messages | JSON |

All shapes built from live `uip is resources describe` output; all six default connections verified Enabled via ping. **Live smoke:** the anthropic flavor ran end to end through the real gateway (`"text": "OK"`). Configurable via `model_spec.is_tools` (`flavors` / `connections` / `models`) — defaults target the `llm_gateway_automated_testing` alpha tenant; other orgs must override `connections`. Note: azure/foundry-backed flavors need connection-specific *deployment* names (a wrong one fails legibly with `DeploymentNotFound`, observed live).

Cost note: cells run per path × per flavor (each ≈2 model calls + 1 IS call) — trim `flavors` or `paths` in `input.json` for cheaper runs.

denispetre:
Superseded by the clean-diff PR: same final state (four payloads incl. call-only is_tools, Studio Web coded-copy agents) without the intermediate churn. See the new PR referenced above.

PR #1011, opened 2026-07-28
feat(chat): forgiving string coercion for vendor_type/api_flavor in get_chat_model

denispetre:
## What

`get_chat_model` documents `vendor_type: VendorType | str | None`, `api_flavor: ApiFlavor | str | None`, and `routing_mode: RoutingMode | str` — but string handling downstream is strict exact-value enum matching:

| input | before | after |
|---|---|---|
| `"awsbedrock"` | ✅ | ✅ |
| `"AWSBedrock"` | ❌ rejected | ✅ `VendorType.AWSBEDROCK` |
| `"CONVERSE"` | ❌ rejected | ✅ `ApiFlavor.CONVERSE` |
| `"anthropic_messages"` | ❌ rejected (must spell `AnthropicMessages`) | ✅ `ApiFlavor.ANTHROPIC_MESSAGES` |
| `"not-a-flavor"` | opaque downstream failure | `ValueError` listing the valid values, raised **before** any factory call |

The wrapper now resolves string inputs case-insensitively against both member **names** and **values**, treating `-`/`_` as interchangeable. Enum members and `None` pass through unchanged — fully backward compatible (only previously-failing inputs change behavior).

## Why

Callers supplying the API flavor as a string (config files, CI workflow inputs — e.g. the model-onboarding testcase's `vendor:flavor` path syntax in #1009) shouldn't need to know that Bedrock's Anthropic flavor is spelled `AnthropicMessages` while everything else is kebab-case.

## Tests

`tests/chat/test_chat_model_factory_coercion.py` — 23 tests: spelling variants per enum, pass-through of members/None, forwarding into the client factory (patched), and the early `ValueError` with valid values listed. Ruff clean; pre-existing collection errors in `tests/chat/` (missing optional `bedrock` extras in a non-`--all-extras` venv) are unrelated and unchanged.

### DIFF
--- a/src/uipath_langchain/chat/chat_model_factory.py
+++ b/src/uipath_langchain/chat/chat_model_factory.py
@@ -10,7 +10,8 @@ passed through as the ``use_new_llm_clients`` argument of :func:`get_chat_model`
   before the ``uipath_langchain_client`` migration.
 """
 
-from typing import Any, Final
+from enum import Enum
+from typing import Any, Final, TypeVar
 
 from langchain_core.callbacks import BaseCallbackHandler, Callbacks
 from langchain_core.language_models import BaseChatModel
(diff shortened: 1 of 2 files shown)


denispetre:
Folded into #1009 per review preference (commit aa36733) — one PR carries the testcase payloads, the workflow flavor input, and the get_chat_model string coercion together.

PR #1012, opened 2026-07-28
fix(agent): align ixp extraction tool with the job attachment strategy

radu-mocanu:
## Summary
- the ixp extraction tool now generates its args schema from `resource.inputSchema` instead of a hardcoded flat model, so it honours the `job-attachment` `$ref` that agent builder already writes
- the tool call resolves job attachments through the shared helper, so the model passes only the attachment id and the full attachment is filled in from state
- an id that is missing from state is reported back to the model instead of reaching the sdk

## Why

the ixp tool was the only attachment-carrying tool not following the shared strategy. it ignored `resource.inputSchema` entirely, hardcoded an alias-free mirror of `Attachment`, and typed the id as `uuid.UUID`.

that last part broke tool simulation: langchain validates arguments against `args_schema` before invoking the tool, so the tool received a live `UUID` object rather than the string the model emitted, and anything encoding those kwargs with a plain `json.dumps` failed with `Object of type UUID is not JSON serializable`.

`process_tool`, `context_tool`, `deeprag_tool` and `analyze_files_tool` all generate the schema from the resource, keep the id a string, and convert to `uuid.UUID` only at the sdk call site. the ixp tool now does the same, which fixes the serialization failure as a side effect of the alignment rather than as a special case.

the arg-resolution half of `get_job_attachment_wrapper` is extracted as `resolve_job_attachment_args` so the ixp wrapper can reuse it while keeping its own result handling (`tools_storage` plus the `dataProjection` tool message that `ixpVsEscalation` depends on).

### DIFF
--- a/src/uipath_langchain/agent/tools/extraction_tool.py
+++ b/src/uipath_langchain/agent/tools/extraction_tool.py
@@ -1,20 +1,29 @@
 """Ixp extraction tool."""
 
-import uuid
-from typing import Any, Optional
+from typing import Any
 
 from langchain.tools import BaseTool
 from langchain_core.messages import ToolCall, ToolMessage
 from langchain_core.tools import StructuredTool
 from langgraph.types import Command, interrupt
-from pydantic import BaseModel, Field
 from uipath.agent.models.agent import AgentIxpExtractionResourceConfig
 from uipath.eval.mocks import mockable
+from uipath.platform.attachments import Attachment
 from uipath.platform.common import DocumentExtraction
 from uipath.platform.documents import ExtractionResponseIXP
 from uipath.platform.errors import EnrichedException
+from uipath.runtime.errors import UiPathErrorCategory
 
-from uipath_langchain.agent.react.job_attachments import raise_for_job_attachment_error
+from uipath_langchain.agent.exceptions import (
+    AgentRuntimeError,
+    AgentRuntimeErrorCode,
+)
+from uipath_langchain.agent.react.job_attachments import (
+    get_job_attachment_paths,
+    get_job_attachments,
+    raise_for_job_attachment_error,
+)
+from uipath_langchain.agent.react.jsonschema_pydantic_converter import create_model
 from uipath_langchain.agent.react.types import AgentGraphState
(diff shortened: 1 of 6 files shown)


src/uipath_langchain/agent/tools/extraction_tool.py
```diff
@@ -29,54 +37,41 @@ class StructuredToolWithWrapper(StructuredToolWithOutputType, ToolWrapperMixin):
...
     project_name = resource.properties.project_name
     version_tag = resource.properties.version_tag
 
+    input_model: Any = create_model(resource.input_schema)
+
     @mockable(
         name=resource.name,
         description=resource.description,
-        input_schema=ExtractionToolInputSchema.model_json_schema(),
+        input_schema=input_model.model_json_schema(),
```
andreitava-uip: Do we know that deployed agents with IXP tools have been migrated? or that the `resource.input_schema` had always been populated?

Worried we might break some old agents for which the schema in `agent.json` was not defined and we relied on the predefined pydantic model

src/uipath_langchain/agent/tools/extraction_tool.py
```diff
@@ -109,7 +104,15 @@ async def extraction_tool_wrapper(
         call: ToolCall,
         state: AgentGraphState,
     ) -> ToolWrapperReturnType:
+        error = resolve_job_attachment_args(tool, call, state)
+        if error is not None:
+            return error
+
         tool_result = await tool.ainvoke(call["args"])
+        # A simulated call returns a validated ExtractionResponseIXP; a real one
```
radu-mocanu: will revert this. tool wrapper doesn t need to know about simulation runtime

src/uipath_langchain/agent/tools/extraction_tool.py
radu-mocanu: the schema is always defined in agent builder. however seems like the CLI is scaffolding an empty schema. 
added for the time being a fallback inout schema to be back compat

andreitava-uip (approved):

PR #1013, opened 2026-07-28
feat: add UiPathByoGuardrailMiddleware (BYOG) [AL-512]

apetraru-uipath:
## What changed?

Adds **Bring Your Own Guardrail (BYOG)** support for coded agents — middleware flavor.

**New middleware**

- **`UiPathByoGuardrailMiddleware`** (`src/uipath_langchain/guardrails/middlewares/byo.py`): runs a customer-managed validator (a cloud content-safety subscription, a vendor validation service, or a custom Integration Service connector) configured under **Admin → AI Trust Layer → Guardrails Configurations**. References the configuration by `validator_name` (`byoValidatorName`) + recommended `connection_id`, and builds a `builtInValidator` guardrail with `validator_type="byo"` — reusing the shared `BuiltInGuardrailMiddlewareMixin` hook wiring unchanged (scopes × stage, tool `match_names`, `asyncio.to_thread` offload, exception routing). No changes to the mixin itself.
- `validator_parameters` is an optional passthrough of raw `ValidatorParameter` values, because BYO parameter schemas are defined by the connector. Scopes and stages are developer-supplied: BYOG validators are **not** scope- or stage-restricted, so all three scopes and both stages are available exactly as for the built-in validators.
- Exports from `uipath_langchain.guardrails` and `.middlewares`.

**Small refactor across existing middlewares** (review feedback)

- Extracted `BUILT_IN_VALIDATOR_GUARDRAIL_TYPE` in `middlewares/_base.py` and used it in **all 7** built-in middlewares in place of the repeated `"builtInValidator"` literal. It is typed `Literal["builtInValidator"]` because `GuardrailType.BUILT_IN_VALIDATOR.value` is typed `str` and does not satisfy the model's `Literal` field under mypy.

**Docs & sample**

- `docs/guardrails.md`: new "Bring Your Own Guardrail (BYOG)" section (admin prerequisite, config lookup, error semantics) plus a "Tuning a BYOG validator" subsection covering `validator_parameters`.
[...]

PR #1014, opened 2026-07-29
feat: support ByoValidator (BYOG) on the decorator path [AL-512]

apetraru-uipath:
## What changed?

Decorator-flavor **Bring Your Own Guardrail (BYOG)** support — the counterpart to `UiPathByoGuardrailMiddleware` shipped in #1013.

**`ByoValidator` re-export**

- Exported from `uipath_langchain.guardrails` and the `.guardrails.decorators` shim as a **plain import**. `uipath-platform` is floored at `>=0.2.14` — the release that ships `ByoValidator` (UiPath/uipath-python#1833, now merged and published) — so the guarded `getattr(..., None)` fallback this PR previously carried is gone, along with the sample's `ImportError` guard and the conditional test skips.
- No new adapter machinery — `@guardrail(validator=ByoValidator("…", connection_id="…"), action=…)` works through the existing `LangChainGuardrailAdapter`, with scope inferred from the decorated target (tool → TOOL, LLM factory → LLM, agent factory → AGENT).

**Adapter hardening**

- The LLM/agent wrappers previously swallowed evaluator exceptions with a bare `except Exception: return` (4 sites in `_langchain_adapter.py`). They now log a WARNING with the guardrail name and traceback while staying fail-open, so a misconfigured BYOG configuration (`PROVIDER_ERROR`, removed/disabled config, BYOG not enabled) is visible instead of silent.

**Parity suite: BYOG scenario (middleware vs decorator)**

- Both parity mock agents (`tests/cli/mocks/parity_agent_{middleware,decorator}.py`) gain the same **"Agent BYOG Detection"** guardrail — AGENT scope, PRE, `BlockAction`; wired through `UiPathByoGuardrailMiddleware` on one side and `ByoValidator` + `@guardrail` on the other, with identical placeholder validator name / connection id.
- New scenario 12, `test_byog_agent_block` in `tests/cli/test_guardrails_in_langgraph.py`, runs for **both flavors** and asserts two parity dimensions: the identical `AgentRuntimeError` block contract on a failing verdict, and — the BYOG-specific part — that both flavors deliver the **same `byo` wire contract** to the guardrails service (`validator_type="byo"`, `byoValidatorName`, `byoConnectionId`). The guardrail objects are captured outside the mock so a mismatch fails the test rather than being swallowed by a fail-open path.

**Docs & sample: ONE sample, both flavors in one agent**

- The two BYOG samples are merged into **`samples/joke-agent-bring-your-own-guardrail`**: a single agent guarded twice from the same BYOG configuration — `UiPathByoGuardrailMiddleware` **logs** on AGENT scope (PRE_AND_POST, `LogAction`) while `ByoValidator` + `@guardrail` on a `create_llm()` factory **blocks** on LLM scope (PRE, `BlockAction`). On a harmful topic you see both actions in sequence: the agent guardrail logs the violation, then the LLM guardrail aborts the run before the model is invoked. `samples/joke-agent-byog-decorator/` is **removed**.
- `docs/guardrails.md`: BYOG section under the decorator pattern. The admin prerequisite and discovery steps are **cross-referenced to the middleware BYOG section** rather than repeated verbatim — only the decorator-specific notes (`parameters` naming, no static stage restriction, error semantics) are spelled out.
- Discovery uses **`uip agent guardrails list --byo`**, whose entries carry `ByoValidatorName` (→ `validator_name`), `ByoConnectionId` (→ `connection_id`) and the full connector-defined `Parameters` schema. Updated in both BYOG doc sections, the sample README and the sample's code comments.
- **No solution binding for the BYOG connection** (review feedback): the connection is resolved server-side from the BYOG configuration, so the sample ships no `bindings.json` and the docs no longer point `ByoConnectorKey` at a connection binding.
- All identifiers are **placeholders** (`my-harmful-content-guardrail`, `my-byog-guardrail-connection`, public `uipath-azure-contentsafety-guardrails` connector key); the Azure Content Safety parameter example ships commented out to keep the sample vendor-neutral.

**Leftovers cleaned up from the #1013 review**

- **Tenant-specific values removed from tests.** The earlier placeholder sweep covered docs and samples but missed the test suites: `tests/guardrails/test_byo_validator.py` and `tests/guardrails/middlewares/test_byo.py` still hard-coded a real validator name and connection GUID. Both now use the same placeholders as the samples. The second file came from #1013 and is already on `main`, so this PR touches it.
- **Scopes/stages claim** and the **`uip agent guardrails list`** discovery command were corrected here in the previous round, matching what landed in #1013.
- **Docs deduplication** after #1013 merged: two `### Bring Your Own Guardrail (BYOG)` sections now exist in one document, so the identical admin-prerequisite and discovery paragraphs were collapsed into a cross-reference.

**Release**

- Version bump to **0.14.20** with `uv.lock` synced (0.14.19 was taken by #1008 and is published on PyPI). The lock also moves `uipath-platform` 0.2.12 → 0.2.14 for the raised floor.

## How has this been tested?

- `tests/guardrails/test_byo_validator.py` — **10 tests, all running, no skips**: adapter evaluator-error logging across all four LLM/agent PRE/POST wrappers (message names the guardrail, traceback attached), and `ByoValidator` through `@guardrail` on a tool, a plain function, and reused across two targets — asserting the `byo` guardrail with `byoValidatorName`/`byoConnectionId` reaches `evaluate_guardrail`. Previously 3 of these skipped on CI behind `skipif ByoValidator is None`; with the floor at `uipath-platform>=0.2.14` they now execute against the published package.
- **Parity suite, scenario 12** — `test_byog_agent_block` runs for BOTH flavors: each parity mock agent gains an "Agent BYOG Detection" guardrail (AGENT scope, PRE, `BlockAction`; middleware via `UiPathByoGuardrailMiddleware`, decorator via `ByoValidator`), and the test asserts they deliver the **same `byo` wire contract** to the service (`validator_type="byo"`, `byoValidatorName`, `byoConnectionId` — captured outside the mock so fail-open paths can't swallow a mismatch) and raise the identical `AgentRuntimeError` block contract. All 12 parity scenarios × 2 flavors green.
- Full suite green; `ruff check`, `ruff format --check`, `mypy` clean; `uv lock --check` clean.
- Verified the re-export resolves without the guard: `ByoValidator` imports from both entry points, is present in both `__all__`s, and builds a guardrail with `validator_type="byo"` carrying the expected `byoValidatorName`/`byoConnectionId`.
- All three `python` fences across both BYOG doc sections compile-checked; the merged sample's `graph.py` parses and its JSON manifests validate.
- **Live E2E on alpha** (`AdminStudioTest`) against a real BYOG harmful-content configuration, running the merged sample (local `uipath-langchain` + published `uipath-platform 0.2.14`):
  - benign topic → both flavors evaluate in one run (agent PRE, LLM PRE per model call, agent POST — all 200) → joke returned ✅
  - harmful topic → `[GUARDRAIL] [BYOG Harmful Content] Failed` logged by the agent-scope middleware, then `Guardrail [BYOG LLM Harmful Content] blocked execution` → `AgentRuntimeError` before `chat/completions` ✅ (vendor verdict: *Harmful content detected: Violence (severity 4)*)

> Behaviour note for reviewers: `@guardrail` on a factory function evaluates the factory's own (empty) arguments once at import time — so a run shows one guardrail call before `START` in addition to the runtime evaluations. This is pre-existing platform behaviour for any factory-decorated guardrail (it also affects the existing `joke-agent-decorator` sample), not something introduced here, and is deliberately left as is. LLM-scope guardrails evaluate once per model call (twice in a two-turn ReAct run).

- [ ] Under Feature Flag
- [x] None
- [ ] DB migrations
- [ ] API removals

Additive: one re-export plus docs, sample and test coverage. Two things a reviewer should note as deliberate: the adapter now emits a WARNING on paths that previously failed silently, and the `uipath-platform` floor moves 0.2.12 → 0.2.14 (within the existing `<0.3.0` range), which is required for `ByoValidator` to exist.


src/uipath_langchain/guardrails/decorators/__init__.py
```diff
@@ -9,9 +10,16 @@
     guardrail,
 )
 
+# Bring Your Own Guardrail (BYOG) validator. Available when the installed
```
valentinabojan: To remove this comment once the TODO is solved

src/uipath_langchain/guardrails/__init__.py
```diff
@@ -43,6 +44,12 @@
     UiPathUserPromptAttacksMiddleware,
 )
 
+# Bring Your Own Guardrail (BYOG) validator. Available when the installed
```
valentinabojan: Same here

docs/guardrails.md
```diff
@@ -452,6 +452,44 @@ async def my_node(state: Input) -> Output:
...
+```bash
+uip agent guardrails byo-configurations --output json
+```
+
+Each entry carries `ValidatorName` and `ConnectionId` (the two values above) plus the resolved connection details (`ConnectorKey`, `ConnectorName`, `ConnectionName`, `ValidConnection`). The command needs an **org-admin user session** — application/client-credentials tokens are rejected — and exits `1` with `Code: ByoGuardrailsUnavailable` when BYOG is not enabled for the tenant. Requires a CLI release that includes the command ([UiPath/cli#3298](https://github.com/UiPath/cli/pull/3298)).
+
+Notes:
+
+- **Always pass `connection_id`.** Validator names are unique per connection only; without the id the server picks the first configuration matching the name.
+- **Stages must match the connector.** BYO validator capabilities are connector-defined, so `ByoValidator` applies no static stage restriction; an unsupported scope/stage surfaces as a `PROVIDER_ERROR` at runtime.
```
valentinabojan: We don't have stage constraints in the connector. We left full flexibility - the connector allows all stages(by not enforcing them in any way) and the guardrail user is in charge to select the stage.

valentinabojan (approved):

docs/guardrails.md
```diff
@@ -242,13 +242,13 @@ byog = UiPathByoGuardrailMiddleware(
...
-To find the `validator_name` and `connection_id`, list the validators available to your tenant:
+To find the `validator_name` and `connection_id`, list your tenant's BYOG configurations:
 
 ```bash
-uip agent guardrails list --output json
+uip agent guardrails list --byo --output json
 ```
 
-Your BYOG configurations are listed alongside the built-in validators — so the same validator type can appear more than once, once for the UiPath-managed validator and once per BYOG configuration. Copy the validator name and connection id from your configuration's entry.
+Each entry carries `ByoValidatorName` (→ `validator_name`), `ByoConnectionId` (→ `connection_id`), and `ByoConnectorKey` (the connector key, useful for a `bindings.json` connection binding), alongside its `AllowedScopes` / `GuardrailStages` and the full `Parameters` schema.
```
valentinabojan: There is no need for the connection details to end up in bindings, for byog.

valentinabojan (approved):

docs/guardrails.md
apetraru-uipath: Done in eb878f6 — the sample no longer ships a `bindings.json` (its only resource was the BYOG connection binding), and the docs/README no longer point `ByoConnectorKey` at a connection binding. Both now state the connection is resolved server-side from the BYOG configuration, with the agent sending only the validator name + connection id.

PR #1016, opened 2026-07-31
feat: support runtime system prompt interpolation in advanced agents

andreitava-uip:
## Summary

- accept callable system prompts in `create_advanced_agent_graph` and resolve them once from invocation inputs
- carry the resolved prompt through dynamically declared graph state and prepend it to each Deep Agents model request
- preserve the prepared Deep Agents prompt content blocks, prompt ordering, and cache metadata
- avoid input-schema collisions by using `uipath__system_prompt` by default and selecting the first available deterministic suffix
- bump `uipath-langchain` to 0.14.20

## Verification

- `uv lock --check`
- Ruff format and lint
- mypy on changed implementation and tests
- targeted advanced-agent and workspace-memory tests (16 passed)
- manual local validation

### DIFF
--- a/src/uipath_langchain/_utils/_pydantic.py
+++ b/src/uipath_langchain/_utils/_pydantic.py
@@ -1,0 +1,42 @@
+"""Shared Pydantic model utilities."""
+
+from pydantic import AliasChoices, AliasPath, BaseModel
+
+
+def get_unique_model_field_name(
+    preferred_name: str,
+    *models: type[BaseModel] | BaseModel | None,
+) -> str:
+    """Return a deterministic field name unused by the supplied Pydantic models."""
+    occupied_names: set[str] = set()
+
+    for model in models:
+        if model is None:
+            continue
+        model_type = model if isinstance(model, type) else type(model)
+        occupied_names.update(model_type.model_fields)
+
+        for field in model_type.model_fields.values():
+            aliases: list[str | AliasPath] = []
+            if field.alias is not None:
+                aliases.append(field.alias)
+            if isinstance(field.validation_alias, AliasChoices):
+                aliases.extend(field.validation_alias.choices)
+            elif field.validation_alias is not None:
+                aliases.append(field.validation_alias)
+
+            for alias in aliases:
+                if isinstance(alias, str):
+                    occupied_names.add(alias)
+                elif alias.path and isinstance(alias.path[0], str):
+                    occupied_names.add(alias.path[0])
+
+    if preferred_name not in occupied_names:
+        return preferred_name
+
+    for suffix in range(1, len(occupied_names) + 1):
+        candidate = f"{preferred_name}_{suffix}"
+        if candidate not in occupied_names:
+            return candidate
+
+    raise AssertionError("A unique model field name must exist")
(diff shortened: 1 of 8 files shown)


src/uipath_langchain/agent/advanced/agent.py
```diff
@@ -27,14 +33,91 @@
...
+
+        if request.system_message is None:
+            system_message = SystemMessage(content=runtime_prompt)
+        else:
+            system_message = SystemMessage(
+                content_blocks=[
+                    {"type": "text", "text": f"{runtime_prompt}\n\n"},
+                    *request.system_message.content_blocks,
+                ]
+            )
```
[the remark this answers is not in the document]
andreitava-uip: Fixed in 2d63991. The middleware now clones the prepared SystemMessage with model_copy and updates only content, preserving id, name, additional_kwargs, response_metadata, and the existing content blocks. Added regression assertions for all of those fields.

src/uipath_langchain/_utils/_pydantic.py
```diff
@@ -0,0 +1,42 @@
...
+    *models: type[BaseModel] | BaseModel | None,
+) -> str:
+    """Return a deterministic field name unused by the supplied Pydantic models."""
+    occupied_names: set[str] = set()
+
+    for model in models:
+        if model is None:
+            continue
+        model_type = model if isinstance(model, type) else type(model)
+        occupied_names.update(model_type.model_fields)
```
radu-mocanu, line 17:
```suggestion
        occupied_names.update(
             model_type.model_fields,
             model_type.model_computed_fields)
```
let s include pydantic computed fields as well here in case `AdvancedAgentGraphState` will define one in the future

src/uipath_langchain/_utils/_pydantic.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def get_unique_model_field_name(
+    preferred_name: str,
+    *models: type[BaseModel] | BaseModel | None,
+) -> str:
+    """Return a deterministic field name unused by the supplied Pydantic models."""
+    occupied_names: set[str] = set()
+
+    for model in models:
```
radu-mocanu, line 13: possible edge case:

if one of the models are defined with `extra="allow"` we cannot check whether the `preferred_name` is taken or not. 
we can be defensive here
```
  if input_schema.model_config.get("extra") == "allow":
      raise ValueError(...)
```

IMO the proper fix would imply defining internal graph state properties at a different level in the graph state
```
{
      "input": {
          "question": "...",
          "uipath_system_prompt": "given by the user",
      },
      "internal": {
          "uipath_system_prompt": "runtime defined system prompt",
      },
  }
```

src/uipath_langchain/_utils/_pydantic.py
andreitava-uip: Addressed in 6135fc2. The allocator now includes computed-field names and aliases in the occupied names, with regression coverage for both.

src/uipath_langchain/_utils/_pydantic.py
andreitava-uip: Added regression coverage in 6135fc2 ensuring the combined advanced graph state does not configure extra="allow" and does not retain undeclared fields. An input schema that explicitly opts into extra="allow" remains a deliberate configuration and is outside this scoped change.

radu-mocanu (approved):

PR #1020, opened 2026-08-03
fix: categorize invalid tool input as user error

radu-mocanu:
## Summary

- categorize dynamic tool argument validation failures as user input errors
- identify the failing tool and preserve schema validation details
- guide users to revise the agent prompt or tool configuration

## Why

Invalid generated tool arguments were reported as unexpected runtime failures instead of actionable user errors.

### DIFF
--- a/src/uipath_langchain/agent/tools/base_uipath_structured_tool.py
+++ b/src/uipath_langchain/agent/tools/base_uipath_structured_tool.py
@@ -9,7 +9,13 @@ from langchain_core.runnables import RunnableConfig
 from langchain_core.tools import StructuredTool
 from langchain_core.tools.base import ArgsSchema, _get_runnable_config_param
 from langchain_core.utils.pydantic import get_fields
-from pydantic import BaseModel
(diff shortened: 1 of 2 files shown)


src/uipath_langchain/agent/tools/base_uipath_structured_tool.py
```diff
@@ -129,6 +138,14 @@ def _parse_input(
...
         return parsed
 
+    def _invalid_input_error(self, error: ValidationError) -> AgentRuntimeError:
+        return AgentRuntimeError(
+            code=AgentRuntimeErrorCode.INVALID_INPUT_ARGUMENT,
+            title=f"Invalid input for tool '{self.name}'",
+            detail=str(error),
+            category=UiPathErrorCategory.USER,
+        )
+
```
andreitava-uip, line 151: nit: User & Deployment errors should be actionable. I would add guidance on what the user can do to solve it, like "revise the prompts and the tool configuration" or something

andreitava-uip (approved):

PR #1025, opened 2026-08-06
feat(mcp): upgrade Python SDK to 2.0

edis-uipath:
## Summary

- upgrade the MCP Python SDK from `1.26.0` to `2.0.0` and replace the copied SDK 1.x Streamable HTTP transport with a thin SDK 2 adapter
- preserve UiPath's externally persisted `SessionInfo`, including restored-session protocol adoption, version headers, idempotent persistence, termination, and recovery from the SDK server's canonical `Session not found` response
- make transport replacement atomic and failure-safe for concurrent callers, while retaining compatibility with the previously accepted `httpx.Timeout` API
- replace the incompatible `langchain-mcp-adapters` sample dependency with a tested first-party active-session tool converter, and migrate local and remote samples to MCP 2 public APIs
- declare a directly tested, bounded `httpx2` dependency and release the public raw-model attribute changes as `uipath-langchain` 0.17.0

## Compatibility

The low-level UiPath client remains compatible with servers negotiating `2025-03-26`, `2025-06-18`, or `2025-11-25`; all three versions are covered by real-transport tests. Externally restored sessions are validated and adopted without another initialize request, then terminated on disposal when configured.

A server that also supports the legacy initialize flow connects in legacy mode. Strictly modern `2026-07-28` discovery-only support remains a separate, uncommitted follow-up layer and is not part of this PR.

## Migration guide

`McpClient` intentionally continues to return the MCP SDK's raw result models. Their Python attributes changed from camelCase to snake_case in SDK 2, so this PR bumps the package to `0.17.0` and documents the exact mappings.

- [Migration guide in this PR](https://github.com/UiPath/uipath-langchain-python/blob/chore/upgrade-mcp-sdk-latest/docs/mcp-sdk-2-upgrade.md)
- [Published documentation route after merge](https://uipath.github.io/uipath-python/langchain/mcp-sdk-2-upgrade/)

## Validation

- `uv lock --check`
- `uv run just lint`
- `uv run ruff format --check .`
- `uv run mypy --config-file pyproject.toml .`
- `uv run pytest` — 2,672 passed
- `uv run pytest -o addopts='' tests/agent/tools/test_mcp` — 79 passed
- real MCP 2 stdio initialization, discovery, and LangChain tool invocation smoke test
- `simple-local-mcp` dependency resolution dry run without `langchain-mcp-adapters`
- `uv build` — 0.17.0 sdist and wheel
- local diff coverage against `origin/main` — 93%

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-langchain==0.16.13.dev1010255648",

  # Any version from PR
  "uipath-langchain>=0.16.13.dev1010250000,<0.16.13.dev1010260000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath-langchain>=0.16.13.dev1010250000,<0.16.13.dev1010260000",
]
```


pyproject.toml
```diff
@@ -21,8 +21,7 @@ dependencies = [
     "openinference-instrumentation-langchain>=0.1.56",
     "jsonschema-pydantic-converter>=0.4.0",
     "jsonpath-ng>=1.7.0",
-    "mcp==1.26.0",
-    "langchain-mcp-adapters==0.2.1",
+    "mcp==2.0.0",
```
radu-mocanu, line 24: Restore compatibility with `langchain-mcp-adapters` before pinning MCP 2.0. The repository’s own `simple-local-mcp` integration resolves `langchain-mcp-adapters==0.3.1` with this pin, then fails at import because the adapter imports `RequestContext` removed by MCP 2. I reproduced the same failure with `uv run --with langchain-mcp-adapters==0.3.1 ...`, and the alpha/cloud/staging integration jobs all fail on it. The adapter metadata only says `mcp>=1.24.0`, so dependency resolution cannot protect downstream users. Land a compatible adapter/migration for this supported path, or prevent the incompatible combination from resolving.

src/uipath_langchain/agent/tools/mcp/streamable_http.py
```diff
@@ -1,802 +1,87 @@
...
+                    url,
+                    http_client=client,
+                    terminate_on_close=terminate_on_close,
+                ) as streams:
+                    yield streams
+        else:
+            async with sdk_streamable_http_client(
+                url,
+                http_client=client,
+                terminate_on_close=terminate_on_close,
```
radu-mocanu, line 82: Preserve the full restored-session lifecycle instead of only injecting its ID into requests. The SDK transport delegated to here keeps a separate in-memory `session_id`; when `SessionInfo` starts with a persisted ID and initialization is skipped, that SDK field remains `None`. I reproduced a successful tool call followed by disposal with `terminate_on_close=True`: no DELETE was sent (`delete_count == 0`) and the persisted ID remained set, whereas the replaced transport terminated through `SessionInfo`. The fresh `ClientSession` also has no adopted negotiation state, so the restored-session request carried no `mcp-protocol-version`. Explicitly hydrate/adopt the required session state or implement termination/version handling for restored IDs, and cover both the protocol header and DELETE path.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -267,38 +278,74 @@ async def _ensure_session(self) -> ClientSession:
...
+                    )
+                    return
+                if self._connection_stack is not None:
+                    await self._connection_stack.aclose()
+                    self._connection_stack = None
+                    self._session = None
                 if self._session_info:
                     await self._session_info.set_session_id(None)
-                await self._initialize_session()
+                await self._open_connection()
```
radu-mocanu, line 310: Keep the client recoverable when opening the replacement connection fails. I reproduced `_open_connection()` raising `MCPError(CONNECTION_CLOSED, ...)` here: `_open_connection` clears `_session`, but `_client_initialized` remains `True`; every later operation then fails in `_ensure_session()` with `RuntimeError("MCP client initialized without a session")` instead of rebuilding the client. The handshake failure also escapes from inside the retry handler, so the configured retry loop does not repair this state. Reset the initialized state on reopen failure or make `_ensure_session()` rebuild when the session is absent, with a focused failed-recovery-then-next-call test.

uv.lock
```diff
@@ -2303,15 +2318,15 @@ wheels = [
 
 [[package]]
 name = "mcp"
-version = "1.26.0"
+version = "2.0.0"
```
radu-mocanu, line 2321: Bump the `uipath-langchain` package version and regenerate the lockfile with that version. This PR changes runtime session/recovery behavior and takes a new major MCP dependency, but `pyproject.toml` and the local-package lock entry still publish `0.15.3`. Every merge publishes immediately, so leaving the package version unchanged makes these changes ride another release and prevents an independent rollback.

PR #1029, opened 2026-08-07
feat: support conversational runtime prompts

radu-mocanu:
## Summary

- resolve advanced conversational system prompts from each exchange input
- preserve custom inputs, user settings, history, aliases, and assistant attachments
- keep shared attachment utilities outside the ReAct package to avoid circular runtime imports

## Why

Advanced conversational agents need the same runtime prompt interpolation as autonomous advanced agents so each turn can resolve its own custom inputs and user context.


src/uipath_langchain/agent/contracts/client_side_tools.py
```diff
@@ -0,0 +1,8 @@
+"""Shared conversational contracts."""
+
+from typing import Any, TypedDict
+
+
+class ClientSideToolInfo(TypedDict):
+    input_schema: dict[str, Any] | None
+    output_schema: dict[str, Any] | None
```
andreitava-uip, line 8: why is this specific to conversation types?

andreitava-uip (approved):

src/uipath_langchain/agent/contracts/client_side_tools.py
radu-mocanu: good catch
I moved it under a generic `contracts` module

PR #1036, opened 2026-08-13
feat(a2a): upgrade Python SDK to v1

PopescuTudor:
## Description

- Bump the uipath-langchain package version from 0.16.3 to 0.16.4.
- Upgrade a2a-sdk from 0.3.x to 1.1.2 and migrate the A2A tool to the v1 protobuf request and response types.
- Negotiate the runtime JSON-RPC protocol from the cached agent card: prefer an advertised v1 interface, otherwise use the SDK's v0.3 compatibility transport for a valid legacy endpoint.
- Populate the SDK client's accepted output modes from the selected card, including the required `acceptedOutputModes` field on v0.3 requests.
- Fail before dispatch when the card advertises neither compatible endpoint; do not retry or downgrade after sending a request.
- Always replace cached third-party URLs with the binding-aware AgentHub proxy URL. The selected SDK transport supplies the matching A2A-Version header and wire format.
- Preserve bearer authentication, task and context continuation, terminal-task cleanup, response extraction, client disposal, and the existing single A2A trace span.
- Prioritize input-required status prompts over stale or partial artifacts.
- AgentHub version-aware proxy routing is already merged in UiPath/AgentHubService#1619.

## Validation

- Full test suite: 2,684 passed (86.20% coverage).
- Focused A2A suite: 27 passed.
- Real Alpha external 0.3-only agent through the AgentHub Remote proxy: verified card negotiation, two `message/send` requests, `A2A-Version: 0.3`, card-derived `acceptedOutputModes`, and non-empty responses.
- Exact published dev wheel against a real Alpha stateful conversational agent: verified both v1 and v0.3 initial requests and continuations with task/context preservation.
- Mypy: passed across 357 source files.
- Ruff lint and format checks: passed across 357 files.
- Custom httpx client lint: passed.
- Wheel and source archive build: passed with package version 0.16.4.
- Vulnerability audit of the changed dependency set: no known vulnerabilities found.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-langchain==0.16.4.dev1010365581",

  # Any version from PR
  "uipath-langchain>=0.16.4.dev1010360000,<0.16.4.dev1010370000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath-langchain>=0.16.4.dev1010360000,<0.16.4.dev1010370000",
]
```

### DIFF
--- a/src/uipath_langchain/agent/tools/a2a/a2a_tool.py
+++ b/src/uipath_langchain/agent/tools/a2a/a2a_tool.py
@@ -12,21 +12,22 @@ import asyncio
 import json
 from contextlib import AsyncExitStack, asynccontextmanager
 from logging import getLogger
-from typing import AsyncGenerator
-from uuid import uuid4
+from typing import Any, AsyncGenerator
+from urllib.parse import urlparse
 
 import httpx
 from a2a.client import Client
+from a2a.helpers import get_artifact_text, get_message_text, new_text_message
 from a2a.types import (
     AgentCard,
+    AgentInterface,
     Message,
-    Part,
     Role,
+    SendMessageRequest,
     Task,
-    TaskArtifactUpdateEvent,
     TaskState,
-    TextPart,
 )
+from google.protobuf.json_format import ParseDict
 from langchain_core.messages import ToolCall, ToolMessage
 from langchain_core.tools import BaseTool
 from langgraph.types import Command
@@ -52,10 +53,10 @@ logger = getLogger(__name__)
 # The A2A terminal task states
 _TERMINAL_TASK_STATES = frozenset(
     {
-        TaskState.completed.value,
-        TaskState.canceled.value,
-        TaskState.failed.value,
-        TaskState.rejected.value,
+        "completed",
+        "canceled",
+        "failed",
+        "rejected",
     }
 )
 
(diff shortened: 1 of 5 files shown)


src/uipath_langchain/agent/tools/a2a/a2a_tool.py
```diff
@@ -103,20 +103,28 @@ async def get(self) -> Client:
...
                         )
-                    self._agent_card.url = agent.a2a_url
+                    runtime_card = AgentCard()
+                    runtime_card.CopyFrom(self._agent_card)
+                    runtime_card.ClearField("supported_interfaces")
+                    runtime_card.supported_interfaces.append(
+                        AgentInterface(
+                            url=agent.a2a_url,
+                            protocol_binding="JSONRPC",
+                            protocol_version="1.0",
```
giuliastf: Are we sure we want to always force A2A 1.0 here? Existing agents that only expose a v0.3 endpoint will start getting 400 Bad Request from the AgentHub proxy. Should we either support both versions or clearly handle the migration for existing agents?

giuliastf (approved):

radugheo (approved):

src/uipath_langchain/agent/tools/a2a/a2a_tool.py
PopescuTudor: Valid concern. AgentHub intentionally rejects a v1 request when the cached card only advertises v0.3, so forcing 1.0 here would break existing v0.3-only bindings. Since a2a-sdk 1.1.2 provides v0.3 client compatibility, I’ll select the runtime interface from the cached card: prefer v1 JSON-RPC when advertised, otherwise use the legacy v0.3 interface, and fail clearly when neither is available. We won’t retry/downgrade after sending a request.

PR #1042, opened 2026-08-19
feat: forward the configured entry point from a process tool

UiPathPetruPopa:
## What changed?

A process tool now runs **the entry point the user selected** rather than whichever one the release happens to default to. `entry_point_path` is read off the tool's stored properties, passed to `invoke_async`, and surfaced in tool metadata — the shape `#1038` established for `folder_key`.

Nothing sets the property yet. The picker that does lives in the Agents repo behind its own flag and ships last.

## The dependency floor is the load-bearing part

`uipath` moves from `>=2.14.6` to `>=2.14.13`, and that is a correctness requirement, not hygiene.

`invoke_async` takes `**kwargs`. An older `uipath` accepts `entry_point_path`, absorbs it, and drops it before the payload is built — so the user's selection is ignored and the job still reports success. Silent success is the exact failure this change exists to remove, and only the floor prevents it.

**2.14.13 is the first published release that declares the field** (checked against the wheels — 2.14.12 does not have it), so the floor is both correct and minimal.

`uipath-platform` needs no change: main already floors it at `0.2.28`, and `0.2.27` was the first release carrying the parameter.

With the floor in place the defensive `getattr` is gone — the field is now guaranteed to exist, so a plain attribute read is the honest one. That was @andreiancuta-uipath's request, and it was only ever blocked on the SDK being unpublished.

## Blast radius

`uv.lock` moves `uipath` 2.14.6 → 2.14.14. That is the only resolved dependency the floor raise moves; nothing else re-pins.

Consumers of `uipath-langchain` inherit the higher `uipath` floor. Anything already on `uipath-platform>=0.2.28` — which main requires — is resolvable.

## How has this been tested?

- 27 tests in `tests/agent/tools/test_process_tool.py` pass. Two added: the stored path reaches `invoke_async`, and it appears in tool metadata.
- **Mutation-checked**: deleting the `entry_point_path=` forward fails four tests, including the three pre-existing exact-args assertions. The forwarding is guarded, not merely executed.
- Those three pre-existing assertions now expect `entry_point_path=None` — direct evidence the default path is unchanged for process, flow and function tools alike.
- `mypy` clean across 392 files; `ruff check` and `ruff format --check` clean; `uv sync --locked` resolves.
- Not tested end-to-end from the UI here — the property has no producer until the Agents-side picker lands.

- [ ] Under Feature Flag
- [x] None
- [ ] DB migrations
- [ ] API removals
- [ ] Authentication/Authorization
- [ ] Deployment pipeline
- [ ] Infrastructure

## Order

The SDK side is merged and published, so nothing upstream blocks this any more. This merges next; the picker UI ships after it, since that is what makes the field user-reachable.


src/uipath_langchain/agent/tools/process_tool.py
```diff
@@ -60,6 +60,9 @@ def create_process_tool(
     # invoke_async accepts only one of folder_path/folder_key, so resolve one.
     folder_path = get_execution_folder_path() or resource.properties.folder_path
     folder_key = get_execution_folder_key() if not folder_path else None
+    # getattr, not attribute access: BaseResourceProperties sets extra="allow", so against a uipath
+    # release predating the declared field the value is present only if the stored JSON carried it.
+    entry_point_path = getattr(resource.properties, "entry_point_path", None)
```
andreiancuta-uipath: issue: You don't need conditional access here, just bump uipath-python once it's merged and access the field normally. It is already typed as Optional.

UiPathPetruPopa: Agreed, but it has to wait: this repo is on uipath 2.14.1, which has no `entry_point_path` on the process properties, so dropping the getattr raises AttributeError today. Will bump the floor and access it normally once #1866 (uipath-python) is merged and published — leaving this open until then.

src/uipath_langchain/agent/tools/process_tool.py
```diff
@@ -95,6 +98,7 @@ async def start_job():
                         parent_span_id=parent_span_id,
                         parent_operation_id=parent_operation_id,
                         run_as_me=True if run_as_me else None,
+                        entry_point_path=entry_point_path,
```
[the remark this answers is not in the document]
UiPathPetruPopa: Correct, and it is the same blocker as the thread above: the floors cannot be raised to a version that does not exist yet. #1866 in uipath-python adds the field; once it merges and publishes, the floors go up in the same change that drops the defensive getattr. Leaving this open until then.

src/uipath_langchain/agent/tools/process_tool.py
UiPathPetruPopa: Done in 84ce0703 — floor raised to `uipath>=2.14.13`, the first release that declares the field, so it is a plain attribute read now. Lock regenerated in 9548ef23.

src/uipath_langchain/agent/tools/process_tool.py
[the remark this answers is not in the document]
UiPathPetruPopa: Floors raised in 84ce0703: `uipath>=2.14.13`, the first release carrying `entry_point_path`. `uipath-platform` needed no change — main already requires 0.2.28. Lock regenerated in 9548ef23.

andreiancuta-uipath (approved):

PR #1044, opened 2026-08-24
fix: never send the solution_folder sentinel as a folder path

Chibionos:
## Problem

`create_process_tool` treats `resource.properties.folder_path` as a real folder path. For **solution-local** resources that value is the literal string `solution_folder` — a packager sentinel meaning *"no folder"* (`SOLUTION_FOLDER_SENTINEL` in the agents repo, `packages/agents-packager/src/bindings/handlers/common.ts`, documented there as *"treat as 'no folder', same as null/empty"*). The packager omits `folderPath` from generated bindings when it sees that value, but never strips it from `agent.json`, so it reaches the runtime.

The fallback chain added in #1038 is:

```python
folder_path = get_execution_folder_path() or resource.properties.folder_path
folder_key  = get_execution_folder_key() if not folder_path else None
```

With `UIPATH_FOLDER_PATH` unset and a solution-local resource, `folder_path` becomes the truthy string `"solution_folder"`, so:

1. `StartJobs` goes out with `X-UIPATH-FolderPath: solution_folder`, a folder that does not exist for the solution; and
2. because it is truthy, `folder_key` is forced to `None`, so `UIPATH_FOLDER_KEY` is never consulted — and the platform client's env-level `FolderContext` header (`_folder_context.py`), which applied before #1038 when `header_folder(None, None)` returned `{}`, is bypassed too.

Net effect: for solution-local tools #1038 **narrowed** the reachable folder set rather than widening it.

Verified in the agents repo that this is a real hazard rather than theoretical: tenants accumulate folders literally named `solution_folder`, `solution_folder 1`, … so a path-based lookup on the sentinel can bind to an unrelated solution's folder.

## Fix

Map the sentinel to `None` before resolving, mirroring the packager's own rule. Precedence is otherwise unchanged: `UIPATH_FOLDER_PATH` → resource `folderPath` → `UIPATH_FOLDER_KEY`.

## Testing

- `test_solution_folder_sentinel_is_not_a_folder_path` — sentinel must not be sent as `folder_path` and must not shadow `UIPATH_FOLDER_KEY`.
- `test_env_folder_path_still_wins_over_sentinel` — `UIPATH_FOLDER_PATH` keeps precedence.
- `tests/agent/tools/test_process_tool.py`: **31 passed**; `ruff check` / `ruff format --check` clean.

## Scope

This is a correctness/hardening fix. It is **not** by itself the fix for the agent-evaluation failures on solution-local workflow tools — in the eval path the shipped snapshot arrives with `folder_path=None`, so this code path is inert there; that root cause (eval jobs not carrying the debug context that lets solution-local workflows resolve) is tracked separately. Filing this on its own because the sentinel-as-folder-path behavior is wrong regardless.

Follow-up noted in #1038 still stands: `context_tool.py`, `escalation_recipient.py`, and `escalation_tool.py` do their own env-path-only resolution and should get the same treatment.

https://claude.ai/code/session_011ZRJcED9mUhXSTkmPSzgbx

### DIFF
--- a/src/uipath_langchain/agent/tools/process_tool.py
+++ b/src/uipath_langchain/agent/tools/process_tool.py
@@ -29,6 +29,9 @@ from uipath_langchain.agent.tools.structured_tool_with_argument_properties impor
 
 from .utils import sanitize_tool_name
 
+# Mirrors SOLUTION_FOLDER_SENTINEL in the agents packager (bindings/handlers/common.ts).
+_SOLUTION_FOLDER_SENTINEL = "solution_folder"
+
 _START_JOBS_ERRORS: dict[tuple[int, str | None], tuple[str, UiPathErrorCategory]] = {
     (404, "1002"): (
         "Could not find process for tool '{tool}'. Please check if the process is deployed in the configured folder.",
@@ -58,7 +61,12 @@ def create_process_tool(
     # Eval serverless jobs expose UIPATH_FOLDER_KEY but not UIPATH_FOLDER_PATH;
     # without a folder header StartJobs returns 404 for a deployed process.
     # invoke_async accepts only one of folder_path/folder_key, so resolve one.
-    folder_path = get_execution_folder_path() or resource.properties.folder_path
+    # "solution_folder" is the packager sentinel for solution-local resources
+    # ("no folder", same as null/empty) — never send it as a real folder path.
+    configured_folder_path = resource.properties.folder_path
+    if configured_folder_path == _SOLUTION_FOLDER_SENTINEL:
+        configured_folder_path = None
+    folder_path = get_execution_folder_path() or configured_folder_path
     folder_key = get_execution_folder_key() if not folder_path else None
 
     input_model: Any = create_model(resource.input_schema)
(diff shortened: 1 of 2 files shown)


src/uipath_langchain/agent/tools/process_tool.py
```diff
@@ -58,7 +61,12 @@ def create_process_tool(
     # Eval serverless jobs expose UIPATH_FOLDER_KEY but not UIPATH_FOLDER_PATH;
     # without a folder header StartJobs returns 404 for a deployed process.
     # invoke_async accepts only one of folder_path/folder_key, so resolve one.
-    folder_path = get_execution_folder_path() or resource.properties.folder_path
+    # "solution_folder" is the packager sentinel for solution-local resources
+    # ("no folder", same as null/empty) — never send it as a real folder path.
+    configured_folder_path = resource.properties.folder_path
+    if configured_folder_path == _SOLUTION_FOLDER_SENTINEL:
+        configured_folder_path = None
```
radu-mocanu, line 68: this check should happen upstream, in the low code package.
if someone builds a coded agent using this method they may end up with unexpected behaviour

radu-mocanu (changes requested):

PR #1049, opened 2026-08-26
feat(deeprag): default to single-shot from-attachments (kill switch: DisableDeepRagFromAttachments)

cfauchere:
## What

Makes DeepRag's internal tool default to the single-call `start_deep_rag_from_attachments` primitive — one durable interrupt returning `CreateDeepRag(attachments=[attachment_id], ...)`; the server creates the ephemeral index and starts the Deep RAG task in one operation. The previous two-step ephemeral-index flow is retained behind a kill switch so we can flip back in production if something goes wrong.

## Changes

- `deeprag_tool.py`:
  - Default path: single `@durable_interrupt` emitting `CreateDeepRag(attachments=[...], index_folder_key=UiPathConfig.folder_key)`.
  - Fallback path (kill switch on): the previous `create_ephemeral_index` + `WaitEphemeralIndex` / `ReadyEphemeralIndex` + `create_deeprag(index_id, index_name, is_ephemeral_index=True)` two-step flow, unchanged.
  - Kill switch: `DisableDeepRagFromAttachments` — declared locally as `DEEP_RAG_FROM_ATTACHMENTS_KILL_SWITCH` because the constant that lived in an earlier draft of the SDK PR never made it into the merged SDK. `FeatureFlags.is_flag_enabled(..., default=False)` resolves it against `UIPATH_FEATURE_DisableDeepRagFromAttachments`.
- `test_deeprag_tool.py`:
  - Default-suite tests cover the new single-shot path — one interrupt, `CreateDeepRag.attachments`, `.prompt`, `.citation_mode`, `.index_folder_key` from `UIPATH_FOLDER_KEY`.
  - Fallback tests set `UIPATH_FEATURE_DisableDeepRagFromAttachments=1` and exercise the two-step path (instant-ready index, wait-for-ingestion, folder-key passthrough).
- `pyproject.toml` / `uv.lock`: bump `uipath-platform>=0.2.20` → `>=0.2.22` for the new primitive; bump package `0.16.12` → `0.16.13`.

## Depends on

- [UiPath/uipath-python#1872](https://github.com/UiPath/uipath-python/pull/1872) — merged. Ships the `start_deep_rag_from_attachments` primitive, `CreateDeepRag.attachments` field, and dispatcher branch this PR relies on.

## Why

Old path: two durable interrupts, one with a `WaitEphemeralIndex` suspend in between while the server ingested the attachments. New path: one interrupt — the server does index creation + ingestion + task start in one call. The trailing wait for the DeepRag task itself is unchanged.

Making the new path the default (rather than opt-in) commits us to the simpler flow; the kill switch is the safety net for a fast rollback without a code change if the server-side path regresses in production.

## Rollback

Set `UIPATH_FEATURE_DisableDeepRagFromAttachments=1` on the runtime. The tool reverts to the previous two-step behaviour immediately — no redeploy needed.

### DIFF
--- a/src/uipath_langchain/agent/tools/internal_tools/deeprag_tool.py
+++ b/src/uipath_langchain/agent/tools/internal_tools/deeprag_tool.py
@@ -9,26 +9,12 @@ from uipath.agent.models.agent import (
     AgentInternalDeepRagToolProperties,
     AgentInternalToolResourceConfig,
 )
-from uipath.core.feature_flags import (
-    DEEP_RAG_FROM_ATTACHMENTS_FEATURE_FLAG,
-    FeatureFlags,
-)
 from uipath.eval.mocks import mockable
-from uipath.platform import UiPath
-from uipath.platform.common import CreateDeepRag, UiPathConfig, WaitEphemeralIndex
-from uipath.platform.context_grounding import (
-    CitationMode,
-    EphemeralIndexUsage,
-)
-from uipath.platform.context_grounding.context_grounding_index import (
-    ContextGroundingIndex,
-)
+from uipath.platform.common import CreateDeepRag, UiPathConfig
+from uipath.platform.context_grounding import CitationMode
 from uipath.runtime.errors import UiPathErrorCategory
 
-from uipath_langchain._utils.durable_interrupt import (
-    SkipInterruptValue,
-    durable_interrupt,
-)
+from uipath_langchain._utils.durable_interrupt import durable_interrupt
 from uipath_langchain.agent.exceptions import AgentStartupError, AgentStartupErrorCode
 from uipath_langchain.agent.react.jsonschema_pydantic_converter import (
     create_model,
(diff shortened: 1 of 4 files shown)


src/uipath_langchain/agent/tools/internal_tools/deeprag_tool.py
```diff
@@ -114,6 +117,22 @@ async def deeprag_tool_fn(**kwargs: Any) -> dict[str, Any]:
             example_calls=[],  # Examples cannot be provided for internal tools
         )
         async def invoke_deeprag(**_tool_kwargs: Any):
+            if not FeatureFlags.is_flag_enabled(
+                DEEP_RAG_FROM_ATTACHMENTS_KILL_SWITCH, default=False
+            ):
+
+                @durable_interrupt
```
radu-mocanu: nit: while this code works, `durrable_interrupt` is not the right primitive here. 
it was created specifically for interrupts in subgraphs. for regular tools we can use the plain `interrupt`.
the PR can be merged as is, we already use `durrable_interrupt` in sibling tools although is not needed

radu-mocanu (approved):

PR #1053, opened 2026-08-28
feat(mcp): support the 2026-07-28 protocol behind an opt-in mode

ionmincu:
Upgrades the MCP Python SDK to 2.0 and adds opt-in support for the `2026-07-28` protocol revision. Supersedes #1025 — this branch now contains that work, so #1025 should be closed rather than merged.

## Why
`McpClient` spoke only the legacy `initialize` handshake, reaching `2024-11-05` through `2025-11-25`. The `2026-07-28` revision replaces that handshake with a stateless `server/discover` probe and drops session IDs entirely. This makes that era reachable while leaving every existing caller's wire behaviour byte-for-byte unchanged.

No high-level client is needed: `ClientSession.discover()` already negotiates `2026-07-28` through UiPath's transport adapter, and the SDK ships the era-negotiation policy `mcp.client._probe.negotiate_auto` that `mcp.Client(mode="auto")` itself uses.

## How agents use it
Nothing changes for existing callers. `create_mcp_tools_and_clients` still defaults to the legacy handshake, and `SessionInfoDebugState` keeps persisting server-minted session IDs exactly as before:
```python
tools, clients = await create_mcp_tools_and_clients(
    resources, session_info_factory=SessionInfoDebugStateFactory(agent_id)
)
```
To opt into the new era, construct the client directly and pass a mode:
```python
client = McpClient(
    config=resource,
    session_info_factory=SessionInfoDebugStateFactory(agent_id),
    protocol_mode="auto",   # probe 2026-07-28, fall back to the handshake
)
```
| `protocol_mode` | Behaviour |
|---|---|
| `"legacy"` *(default)* | `initialize` only — today's behaviour, unchanged |
| `"auto"` | probe `server/discover`, fall back to the handshake |
| `"modern"` | `server/discover` only, no fallback |

**AgentHub needs no change.** `2026-07-28` removes `mcp-session-id`, which AgentHub routes serverless MCP instances by — so in modern mode the client mints its own ID and keeps sending it on that same header as an opaque routing key. A modern server ignores it. Because the client mints it *before* negotiating, it is on the very first request (`server/discover` included), which a server-assigned session ID never could be.

**`uipath-agents-python` needs no change either.** `SessionInfo` stores whichever ID the server uses, so `SessionInfoDebugState` persists a client-minted affinity ID unmodified, and a later run returns to the same warm instance. Verified against a dev build of this branch in UiPath/uipath-agents-python#702: all 36 checks pass.

Known gap: `create_mcp_tools_and_clients` has no `protocol_mode` parameter yet, so the new era is only reachable by constructing `McpClient` directly. Deliberate for now — the default must stay legacy — but worth a follow-up if AgentHub wants it plumbed through.

## What changed
**`ProtocolStrategy` abstracts the session lifecycle, not the transport.** Negotiation is one call in either era; what differs is how a connection is negotiated, whether a restored one can be reused, which errors a reconnect can fix, and how the connection is identified on the wire.

**The default stays `"legacy"` deliberately.** `"auto"` would silently move any discovery-capable UiPath MCP server to stateless `2026-07-28` and stop issuing session IDs, breaking the AgentHub playground persistence `SessionInfoDebugState` exists for.

**Fixes a silent protocol downgrade on resume.** The old code probed each handshake version with a ping and adopted the first that answered — but servers don't validate that header against what the session negotiated, so the oldest always won. A session negotiated at `2025-11-25` was adopted as `2025-03-26`, disabling the server's `2025-11-25` SSE resumability for the rest of the run. The version can't be recovered from the wire (responses carry only the session ID), so resume now re-runs `initialize` *inside* the restored session, which is safe because the server routes by the session header and mints a new session only when that header is absent. Resume drops from up to four round trips to one.

**Two bugs found and fixed while building this**: a proxy echoing `mcp-session-id` back could overwrite the client-minted routing key mid-connection; and disposal sent `DELETE` for a session the server never issued, reaching the gateway as a teardown for the live instance the ID exists to pin — on every run after the first.

**Retries are scoped per era.** A legacy session can be lost and re-established; every modern request is self-contained, so only a dropped connection is retryable there.

Also: `streamable_http.py` drops from ~800 lines to a thin adapter over the upstream transport; `langchain-mcp-adapters` is replaced by a first-party session-to-LangChain converter (it imports `RequestContext`, removed in MCP 2); `httpx.Timeout` remains accepted.

## Testing
`McpClient` is tested against a real `MCPServer` over real HTTP, not a mocked transport — negotiation in every mode, resume asserting the *originally negotiated version*, affinity pinning across clients, per-era retry and disposal, and all four handshake versions (`2024-11-05` and `2025-03-26` were previously untested anywhere). Mocked transports remain only for what a cooperative server can't produce: concurrency races, pathological servers, and pure-function matrices.

`testcases/simple-http-mcp` adds an LLM-free integration testcase hosting MCP over Streamable HTTP on real sockets, covering the same matrix plus the exact API surface `uipath-agents-python` depends on.

## Breaking changes
Version bumped to **0.17.0**. SDK 2.0 renamed the raw result-model attributes `McpClient` deliberately returns:
| Before | After |
|---|---|
| `CallToolResult.isError` | `is_error` |
| `CallToolResult.structuredContent` | `structured_content` |
| `Tool.inputSchema` | `input_schema` |
| `Tool.outputSchema` | `output_schema` |
| `ListToolsResult.nextCursor` | `next_cursor` |

The wire format is unchanged — these became snake_case fields carrying the old names as serialization aliases, so `model_dump(by_alias=True)` still emits camelCase. Plain `model_dump()` now emits snake_case keys, the one form of this break that fails silently rather than raising.

`McpClient.SESSION_ERROR_CODES` is removed — use the `McpClient.is_session_error` static method. Removing the vendored transport also drops the unused public names it carried: `streamablehttp_client`, `StreamableHTTPTransport`, `RequestContext`, `StreamableHTTPError`, `ResumptionError`.

One thing to flag: modern mode sends `mcp-session-id`, a header `2026-07-28` doesn't define for requests, purely as a routing key. The SDK server ignores it; a stricter server or proxy could object. Worth confirming with whoever owns UiPath's MCP server implementation.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-langchain==0.17.0.dev1010535708",

  # Any version from PR
  "uipath-langchain>=0.17.0.dev1010530000,<0.17.0.dev1010540000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath-langchain>=0.17.0.dev1010530000,<0.17.0.dev1010540000",
]
```


src/uipath_langchain/agent/tools/mcp/protocol_strategy.py
```diff
@@ -0,0 +1,295 @@
...
+            await session.initialize()
+            logger.info(
+                "MCP session initialized with session ID: %s",
+                await info.get_session_id(),
+            )
+            return
+
+        try:
+            result = await session.initialize()
+        except MCPError as error:
```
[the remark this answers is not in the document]
ionmincu: Fixed in a91aeed — good catch, this was a real bug.

`CONNECTION_CLOSED` now re-raises without touching the stored ID. The consequence you spotted is worse than an unnecessary drop: for a store-backed `SessionInfo` like `SessionInfoDebugState`, `reset()` **writes** `None` to the debug-state endpoint, so a transient blip during resume destroyed the persisted session permanently — and the follow-up `initialize()` ran on the same dead transport and failed anyway. Recovery now reopens and resumes instead.

Guarded by `test_legacy_keeps_a_persisted_session_when_the_connection_drops`, verified to fail without the fix.

src/uipath_langchain/agent/tools/mcp/protocol_strategy.py
```diff
@@ -0,0 +1,295 @@
...
+        # the legacy wire before the era is known: a modern server simply never
+        # sends one back, leaving nothing to capture.
+        self.identity = SessionIdentity(LEGACY_IDENTITY)
+        self._legacy = LegacyHandshakeStrategy()
+        self._modern = ModernDiscoveryStrategy(affinity_meta_key)
+        self._resolved: ProtocolStrategy = self._legacy
+
+    async def connect(self, session: ClientSession, info: SessionInfo) -> None:
+        """Negotiate an era, then narrow this strategy to it."""
+        self.identity.wire = LEGACY_IDENTITY
```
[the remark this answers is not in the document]
ionmincu: Fixed in a91aeed — correct, and reachable.

`_resolved` is now reset to legacy at the top of `connect()`, alongside the existing wire reset. The path you describe is real: `_reinitialize_session` → `_open_connection` → `connect()` raises inside `negotiate_auto`, then `_execute_with_retry` consults `is_recoverable` with the stale modern policy, which refuses to retry session errors — so a legacy server reached on the retry would never recover.

Guarded by `test_auto_does_not_carry_a_stale_era_through_a_failed_probe`, verified to fail without the fix.

src/uipath_langchain/agent/tools/mcp/streamable_http.py
```diff
@@ -63,20 +172,22 @@ async def streamable_http_client(
...
+            _inject_meta(request, wire.meta_key, session_id)
 
     async def capture_session_id(response: httpx2.Response) -> None:
         nonlocal sdk_session_id
-        session_id = response.headers.get(MCP_SESSION_ID)
+        capture_header = session_identity.wire.capture_response_header
+        if capture_header is None:
+            return
+        session_id = response.headers.get(capture_header)
         if session_id is not None:
```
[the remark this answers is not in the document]
ionmincu: Fixed in a91aeed. Persistence is now gated on the request being `initialize`, which is exactly what the SDK transport does itself — it calls `_maybe_extract_session_id_from_response` only under `_is_initialization_request` (`mcp/client/streamable_http.py:375`). Matching that is the strongest argument for the change.

Worth noting the window is slightly wider than described: it is not only the first probe. On any later `auto` run the persisted affinity ID is sent on the legacy wire during the probe, so an echoed header could replace it before the era resolves and narrows `capture_response_header` to `None`.

Guarded by `test_only_an_initialize_response_assigns_a_session_id`, using an endpoint that stamps a different ID on every non-initialize response. Verified to fail without the fix.

tests/agent/tools/test_mcp/claude.md
```diff
@@ -44,13 +45,27 @@ tests/agent/tools/test_mcp/
...
 │
+├── test_protocol_strategy.py  # Modern/auto eras + affinity identity
+│   ├── EraMcpEndpoint         # Serves either era, or both
+│   ├── test_modern_mode_negotiates_through_discovery
+│   ├── test_modern_mode_pins_an_affinity_id_from_the_first_request
+│   ├── test_modern_mode_reuses_a_persisted_affinity_id
+│   ├── test_modern_mode_does_not_retry_a_session_shaped_error
+│   ├── test_auto_mode_resolves_to_the_modern_era
+│   ├── test_auto_mode_falls_back_to_the_legacy_handshake
+│   ├── test_auto_mode_sends_a_restored_id_on_both_headers_before_resolving
```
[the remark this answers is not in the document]
ionmincu: Already resolved — this doc was rewritten after the review ran, and `test_auto_mode_sends_a_restored_id_before_the_era_resolves` is the name now listed.

Prompted by this I swept every `test_*` name in the file against the actual test functions and found one genuine stale reference that predates this branch: `test_returns_empty_for_agent_without_mcp`, which does not exist. Corrected to `test_returns_empty_for_empty_resources` in a91aeed.

src/uipath_langchain/agent/tools/mcp/protocol_strategy.py
```diff
@@ -0,0 +1,310 @@
...
+        # Widen back to the conservative era first: if the probe raises, the
+        # previous connection's resolution must not decide how this failure is
+        # recovered from.
+        self._resolved = self._legacy
+        await negotiate_auto(session)
```
radu-mocanu: We cannot leave auto like this: `negotiate_auto()` sends `server/discover` before the affinity ID is minted below. I reproduced discovery on instance-1 and the first tool call on instance-2, so this defeats the serverless pinning goal.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -262,44 +299,80 @@ async def _ensure_session(self) -> ClientSession:
...
+                if failed_session is not None and self._session is not failed_session:
+                    logger.debug(
+                        "MCP session was already replaced by another operation"
+                    )
+                    return
+                await self._close_connection_for_recovery()
                 if self._session_info:
-                    await self._session_info.set_session_id(None)
-                await self._initialize_session()
+                    await self._strategy.reset(self._session_info)
```
radu-mocanu, line 356: Do not reset on every recovery. `CONNECTION_CLOSED` only means the transport dropped; clearing the persisted ID here makes the retry start a new server session. Keep it unless the server actually rejected or terminated it.

src/uipath_langchain/agent/tools/mcp/protocol_strategy.py
```diff
@@ -0,0 +1,310 @@
...
+abstracts.
+"""
+
+import logging
+from dataclasses import replace
+from typing import Literal, Protocol, runtime_checkable
+from uuid import uuid4
+
+from mcp import ClientSession
+from mcp.client._probe import negotiate_auto
```
radu-mocanu: Please do not import `mcp.client._probe`. It is private SDK surface; use a public API or own this small policy here.

src/uipath_langchain/agent/tools/mcp/streamable_http.py
```diff
@@ -1,802 +1,264 @@
...
-            logger.warning(f"Session termination failed: {exc}")
-
-    async def get_session_id(self) -> str | None:
-        """Get the current session ID."""
-        return await self._session_info.get_session_id()
+    ``ByteStream`` is imported here rather than at module scope so this private
+    HTTPX detail cannot break importing the package. Mirroring into ``_meta`` is
+    opt-in, so only that feature depends on it.
+    """
+    from httpx2._content import ByteStream
```
radu-mocanu: Please do not import `httpx2._content.ByteStream` either. This makes the opt-in affinity-meta path depend on HTTPX internals; use a public request/body seam.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
ionmincu: We do not call strategy reset on `connection_Closed` error.

src/uipath_langchain/agent/tools/mcp/protocol_strategy.py
ionmincu: Fixed, now we set affinity id before the probe, so `server/discover` will land on same instance as `tool/calls`

src/uipath_langchain/agent/tools/mcp/protocol_strategy.py
ionmincu: Removed probe and implemented our own.

src/uipath_langchain/agent/tools/mcp/streamable_http.py
ionmincu: We used it to add affinity in meta json, but i dropped it, we only send it now at header level.

radu-mocanu (approved):

PR #1054, opened 2026-08-31
chore(deps): migrate to deepagents 0.7.11

radu-mocanu:
## Summary
- bump deepagents to 0.7.11, raising the langchain floor to 1.3.18 and langgraph to 1.2.11
- offer the QuickJS code interpreter to advanced agents, behind the optional `code-interpreter` extra
- drop the `write_todos` tool, following the benchmark deepagents 0.7.0 based its own removal on
- flag tools that suspend the run, so callers can tell which ones are unsafe to invoke outside the graph's tool node

## Code interpreter

`build_code_interpreter_middleware(tools)` returns middleware giving the agent one `eval` tool: a persistent JavaScript REPL in a WASM guest, with eligible tools callable from inside it as `tools.<camelCaseName>(...)`. Pass it via `middleware=` on either advanced graph builder.

It is opt-in twice over. The dependency is an extra (`uipath-langchain[code-interpreter]`), because wasmtime and the QuickJS guest are ~27 MB on disk and no consumer should carry that unless it asks. The middleware import is deferred into the factory, so importing the package without the extra keeps working and only the factory raises, with install guidance.

The allowlist governing what the sandbox can reach is derived, never hand-written. Three exclusions: tools that suspend the run (a replayed node re-runs every bridged call made before the interrupt, and PTC bypasses approval hooks), names that cannot be JavaScript identifiers (upstream raises mid-turn for those), and camelCase collisions (upstream dedupes by tool name, so binding either would call the wrong tool).

`mode` is a parameter rather than a constant. `"thread"` snapshots the REPL into the checkpoint, so state survives a suspend and resume, at a fixed ~1.25 MB per checkpoint thread written even when the agent never calls `eval`. That only pays off where a later run reads it back, so a caller whose runs each get their own thread should pass `"turn"`, which writes nothing. Upstream renders the tool description from the mode, so the model is told the correct semantics either way.

## `write_todos` removed

deepagents 0.7.0 dropped `TodoListMiddleware` from its defaults after benchmarking across three models: no statistically significant accuracy gain on any of them, and higher token use on two ([langchain-ai/deepagents#4929](https://github.com/langchain-ai/deepagents/pull/4929)). This follows that rather than restoring it, and pins the absence with a test so a future change has to be deliberate.

Kept as its own commit, so it can be reverted without touching the version bump.

## Breaking

`BackendFactory` is gone from `uipath_langchain.agent.advanced`, and the `backend` parameters no longer accept a factory. deepagents 0.7.0 deleted the type, so there is nothing left to alias it to and a factory would be rejected downstream regardless. Pass a `BackendProtocol` instance instead.

Advanced agents no longer expose `write_todos`. A caller that wants it can pass `middleware=[TodoListMiddleware()]`; the middleware injects its own usage prompt, so no caller prompt needs to describe the tool.

Two behaviour changes come from deepagents itself and are worth knowing before upgrading: `write_file` now replaces an existing file rather than erroring, and a recursive `delete` tool appears whenever the backend supports it. `delete` is deliberately not withheld: an agent can already replace a file with `write_file`, and letting it clean up scratch files reduces orphan job attachments.

## Why

deepagents 0.5.9 is two minor versions behind, and 0.7.x is the floor `langchain-quickjs` requires, so the code interpreter was not reachable before this.

the base prompt goes with it: 0.7.0 empties `BASE_AGENT_PROMPT`, which is where its input-token reduction comes from, so this adopts the lean prompt rather than passing the deprecated constant back in.

### DIFF
--- a/src/uipath_langchain/agent/advanced/agent.py
+++ b/src/uipath_langchain/agent/advanced/agent.py
@@ -6,9 +6,8 @@ from typing import Any, Literal, NotRequired, cast
 
 from deepagents import CompiledSubAgent, SubAgent
 from deepagents import create_deep_agent as _create_deep_agent
-from deepagents.backends import BackendProtocol
-from deepagents.backends.filesystem import FilesystemBackend
-from deepagents.backends.protocol import BackendFactory
+from deepagents.backends import BackendProtocol, FilesystemBackend
+from deepagents.middleware.subagents import GENERAL_PURPOSE_SUBAGENT
 from langchain.agents.middleware import (
     AgentMiddleware,
     AgentState,
@@ -27,6 +26,7 @@ from uipath.core.chat import UiPathConversationMessageData
 from uipath.runtime.errors import UiPathErrorCategory
 
 from uipath_langchain._utils import get_unique_model_field_name
+from uipath_langchain.agent.attachments.constants import OUTPUT_FILE_TOOL_NAME
 from uipath_langchain.agent.attachments.job_attachments import get_job_attachment_paths
 from uipath_langchain.agent.attachments.output_files import (
     DEFAULT_MAX_OUTPUT_FILE_RETRIES,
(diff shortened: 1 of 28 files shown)


src/uipath_langchain/agent/advanced/agent.py
```diff
@@ -141,12 +141,69 @@ def _resolve_runtime_system_prompt(
     return _RuntimeSystemPrompt(None, system_prompt, state_key)
 
 
+# Tools the main agent must hold alone. A deepagents subagent runs in its own
+# context and returns a single text report, so an attachment it uploads is real but
+# its reference never reaches the main agent -- the only agent that fills the typed
+# output. Observed result: two orphan attachments and a faulted job.
+MAIN_AGENT_ONLY_TOOLS: frozenset[str] = frozenset({OUTPUT_FILE_TOOL_NAME})
```
radu-mocanu, line 210: note: this implementation will change. subagents will have access to the `generate_attachment` tool (since they can pass a file as input to other tools). will be done in a subsequent PR

PR #1055, opened 2026-09-01
feat: bump jsonschema-pydantic-converter to 0.4.1, add datamodel-code-generator backend

vldcmp-uipath:
Two changes, one dependency apart.

## 1. fix parsing of inline object - fixed by the converter bump

An object written inline rather than under `$defs` that held a `$ref` produced a child model which was never completed: the root reported `__pydantic_complete__ = True` while the child kept a `ForwardRef('__Dynamictype_N')`. It surfaced as

```
X is not fully defined; you should define __Dynamictype_N, then call X.model_rebuild()
```

once LangChain copies the field annotations into a model in another module to build the tool definition — so an **agent failed at tool-binding time, not on execution**.

`jsonschema-pydantic-converter` fixed that upstream in **0.4.1**. The only thing needed here is the pin:

```diff
-    "jsonschema-pydantic-converter>=0.4.0",
+    "jsonschema-pydantic-converter>=0.4.1",
```

`TestInlineObjectWithRef` asserts the shape against **both** backends, so the guarantee cannot regress on either path.

## 2. A second schema backend behind a feature flag

Independent of the fix above. Schemas are converted to Pydantic models by two interchangeable backends now, selected by the `EnableDatamodelCodeGeneratorConverter` feature flag:

| flag | backend |
|---|---|
| **off (default)** | `jsonschema-pydantic-converter` |
| **on** | `datamodel-code-generator` |

`create_model` and `create_output_model` keep their signatures, so **no caller changes** — all 11 call sites here (and the 2 in `uipath-agents-python`) go through the same façade.

```
jsonschema_pydantic_converter.py            façade: flag dispatch + create_output_model
├── _legacy_converter.py                    jsonschema-pydantic-converter
├── _datamodel_code_generator_converter.py  datamodel-code-generator
├── _datamodel_code_generator_base.py       shared base model (serialize_by_alias, extra)
└── _schema_refs.py                         $ref helpers, backend-agnostic
```

The flag is a gradual rollout, **not** an escape from a broken backend — with 0.4.1 in place both resolve `$ref`s completely. It exists for what the newer one does better:

- generated types are named after the schema instead of `DynamicType_N`, and **those names reach the language model** through `$defs`
- every generated class is homed in the conversion's own pseudo-module. The legacy wrapper only re-homes types collected from `$defs`, so a model built for an inline object keeps the converter's *shared* module — where repeated class names across schemas can make a qualified-name lookup land on another schema's class
- property names that are not valid Python identifiers are repaired and carry an alias, so the declared names still go on the wire (the legacy backend emits a field literally named `my-field`)

## Kept equivalent where callers depend on it

- **Original JSON property names on the wire**, via `serialize_by_alias` from a shared base class. `additionalProperties: false` still merges over that base to forbid extras.
- **`__uipath_marker_name__`** on types reached through a `$ref`, byte-compatible with the legacy names because job-attachment discovery looks types up by them (`get_json_paths_by_type(model, "__Job_attachment")`). The mapping comes from walking the schema alongside the model tree — the generator sanitizes and de-duplicates class names on its own terms, so no name-based mapping is reliable.
- **`not`, `prefixItems` and an empty `enum`** have no Pydantic annotation, so the code generator drops them and the field would accept anything. The legacy backend enforces them itself, so those paths are checked against the original subschema with `jsonschema` rather than letting validation silently lapse.
- **`AgentStartupError` naming the unresolved type**, raised before generation so the message stays the one users already see.

## Differences that remain

Measured, not assumed, and pinned in `TestBackendDifferences`:

| | legacy 0.4.1 | datamodel-code-generator |
|---|---|---|
| inline object holding a `$ref` | complete | complete |
| inline model `__module__` | `_type_converters` (shared) | own pseudo-module |
| `$defs` names the model sees | `DynamicType_0/1` | `Fields`, `Project` |
| `additionalProperties`-only map | nested model | `dict[str, str]` |
| root title | `Create_Issue` | `Create_Issue` |

## Testing

The **whole conversion contract suite runs against both backends** — that is what makes this a genuine toggle rather than two divergent paths.

- **2860 passed, 3 skipped, 0 failed**
- 138 tests across the two converter modules, **66 per backend**; the only 6 not parametrized by backend are the ones that switch backends themselves, to compare the two or to assert the flag routing
- `ruff check`, `ruff format --check`, `mypy` clean (src and the touched test packages)
- ~7 ms/schema on the code-generator path (≈280 ms for a 40-schema agent startup)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-langchain"
-version = "0.17.4"
+version = "0.17.5"
 description = "Python SDK that enables developers to build and deploy LangGraph agents to the UiPath Cloud Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
@@ -20,7 +20,9 @@ dependencies = [
     "httpx>=0.27.0",
     "httpx2>=2.5.0, <2.10.0",
     "openinference-instrumentation-langchain>=0.1.69, <0.2.0",
-    "jsonschema-pydantic-converter>=0.4.0",
+    "datamodel-code-generator>=0.76.0",
+    "jsonschema-pydantic-converter>=0.4.1",
+    "jsonschema>=4.23.0",
     "jsonpath-ng>=1.7.0",
     "mcp==2.0.0",
     "pillow>=12.1.1",
@@ -89,6 +91,7 @@ dev = [
     "numpy>=1.24.0",
     "pytest_httpx>=0.35.0",
     "rust-just>=1.39.0",
+    "types-jsonschema>=4.23.0",
     "types-protobuf<7",
     "packaging>=24.0",
     # tests/agent/tools/test_mcp/real_server.py hosts real MCP servers over real
--- a/src/uipath_langchain/agent/react/_datamodel_code_generator_base.py
+++ b/src/uipath_langchain/agent/react/_datamodel_code_generator_base.py
@@ -1,0 +1,60 @@
+"""Base model for every runtime-generated schema class.
+
+`datamodel-code-generator` emits classes deriving from a configurable base. Every
+model generated from a tool or agent schema derives from
+:class:`UiPathDatamodelCodeGeneratorBaseModel`, which supplies the two
+configuration options the runtime depends on:
+
+* ``serialize_by_alias`` -- properties whose JSON names are not valid Python
+  identifiers are generated as sanitized fields carrying an alias. Serializing by
+  alias is what puts the original JSON names back on the wire, so a tool call
+  reaches Integration Service with the property names its schema declared.
+* ``extra="allow"`` -- the default for a schema that does not say otherwise. A
+  schema with ``additionalProperties: false`` generates its own
+  ``model_config``, and Pydantic merges that over this one, so ``extra`` still
+  ends up ``"forbid"`` there.
+
+The base also makes a declared JSON property name usable as an attribute, which
+is the half of that contract sanitizing the field name would otherwise break.
+See :meth:`UiPathDatamodelCodeGeneratorBaseModel.__getattr__`.
+"""
+
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict
+
+
+class UiPathDatamodelCodeGeneratorBaseModel(BaseModel):
+    """Base class for models generated from JSON Schema at runtime."""
+
+    model_config = ConfigDict(serialize_by_alias=True, extra="allow")
+
+    def __getattr__(self, name: str) -> Any:
+        """Resolve a declared JSON property name to its sanitized field.
+
+        Serializing by alias makes ``model_dump()`` alias-keyed, and consumers read
+        the dumped keys straight back off the instance: LangChain's
+        ``BaseTool._parse_input`` builds its kwargs with ``getattr(result, key)``
+        for every dumped key. Under the legacy backend the JSON name *was* the
+        field name, so that resolved; here the field is sanitized, so a property
+        declared ``Content-Type`` would raise ``AttributeError`` mid-tool-call.
+
+        Accepting the alias restores what callers already relied on -- the name the
+        schema declared works as an attribute -- so alias-unaware consumers behave
+        as they did before.
+
+        Only reached when normal lookup fails, so real fields, methods and extras
+        are untouched.
+        """
+        try:
+            return super().__getattr__(name)  # type: ignore[misc]
+        except AttributeError:
+            pass
+
+        for field_name, field in type(self).model_fields.items():
+            if field.alias == name and field_name != name:
+                return getattr(self, field_name)
+
+        raise AttributeError(
+            f"{type(self).__name__!r} object has no attribute {name!r}"
+        )
--- a/src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
+++ b/src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
@@ -1,0 +1,501 @@
+"""Schema-to-model conversion via ``datamodel-code-generator``.
+
+The newer backend. Compared with :mod:`._legacy_converter` it names generated
+types after the schema rather than ``DynamicType_N`` -- those names reach the
+language model through ``$defs`` -- homes every class it produces in this
+conversion's pseudo-module, and repairs property names that are not valid Python
+identifiers while keeping the declared names on the wire.
+
+Both backends resolve ``$ref``s completely.
+
+Everything below the two public entry points exists to keep three runtime
+contracts intact:
+
+* generated classes live in a per-conversion pseudo-module registered in
+  ``sys.modules``, so qualified-name lookups resolve (LangGraph checkpoint
+  deserialization, and :mod:`uipath_langchain.agent.attachments.pydantic_json`);
+* classes reached through a ``$ref`` carry ``__uipath_marker_name__`` holding the
+  name that ``$defs`` entry maps to, which is how job-attachment discovery finds
+  them;
+* schemas whose ``$ref`` targets are missing fail at startup with a message
+  naming the type, rather than producing a model that breaks later.
+"""
+
+import itertools
+import keyword
+import re
+import sys
+from collections.abc import Iterator
+from types import ModuleType
+from typing import Any, Type, cast, get_args
+
+import jsonschema
+from datamodel_code_generator import (
+    Formatter,
+    GenerateConfig,
+    InputFileType,
+    generate_dynamic_models,
+)
+from datamodel_code_generator.parser.jsonschema import json_schema_data_formats
+from pydantic import BaseModel, model_validator
+
+from uipath_langchain.agent.exceptions import AgentStartupError, AgentStartupErrorCode
+
+from ._datamodel_code_generator_base import UiPathDatamodelCodeGeneratorBaseModel
+from ._schema_refs import ref_resolves, resolve_pointer
+
+# Prefix for the per-conversion pseudo-modules that let qualified-name lookups
+# resolve each schema's generated classes.
+_DYNAMIC_MODULE_PREFIX = "jsonschema_pydantic_converter._dynamic"
+
+_dynamic_module_counter = itertools.count()
+
+# Import path of the base class every generated model derives from. The generator
+# takes this as a string and emits the import itself.
+_BASE_CLASS = (
+    f"{UiPathDatamodelCodeGeneratorBaseModel.__module__}"
+    f".{UiPathDatamodelCodeGeneratorBaseModel.__name__}"
+)
+
+# Pin every ``format`` back to its type's own default, so a format annotates the
+# value without changing it. Left alone, the generator retypes on format --
+# ``password`` to ``SecretStr``, which serializes as a mask rather than the
+# credential; ``email`` and ``ulid`` to types whose validators are not installed,
+# which fails conversion outright; ``date-time`` to a timezone-aware datetime,
+# rejecting the naive stamps models emit; and a dozen more to objects that are no
+# longer JSON-serializable once dumped. None of that is what this backend is for,
+# and the legacy backend does none of it, so the two must agree here.
+#
+# Derived from the generator's own table rather than listed, so a format added
+# upstream is covered without a change here.
+_FORMAT_PINS = [
+    f"{base}+{name}={base}"
+    for base, formats in json_schema_data_formats.items()
+    for name, type_ in formats.items()
+    if type_ is not formats.get("default")
+]
+
+# JSON Schema keywords that have no equivalent in Pydantic's type system, so the
+# generator drops them and the field ends up accepting anything. Enforcement for
+# these is restored by _build_constraint_guard.
+_UNENFORCEABLE_KEYWORDS = ("not", "prefixItems")
+
+# Keywords that nest a subschema describing the same value as their parent.
+_COMBINERS = ("anyOf", "oneOf", "allOf")
+
+_INVALID_SCHEMA_TITLE = "Invalid schema"
+
+
+def _invalid_schema(detail: str) -> AgentStartupError:
+    """The startup error raised for every schema this backend cannot convert."""
+    return AgentStartupError(
+        code=AgentStartupErrorCode.INVALID_TOOL_CONFIG,
+        title=_INVALID_SCHEMA_TITLE,
+        detail=detail,
+    )
+
+
+def _create_dynamic_module() -> ModuleType:
+    """Create a pseudo-module unique to one conversion, since class names repeat."""
+    module_name = f"{_DYNAMIC_MODULE_PREFIX}_{next(_dynamic_module_counter)}"
+    pseudo_module = ModuleType(module_name)
+    sys.modules[module_name] = pseudo_module
+    return pseudo_module
+
+
+def _definition_type_name(ref: str) -> str:
+    """The name a ``$ref`` maps to, byte-compatible with the legacy backend."""
+
+    def sanitize(name: str) -> str:
+        return re.sub(r"[^a-zA-Z0-9_]", "_", name)
+
+    if ref.startswith("#/"):
+        parts = [
+            sanitize(part)
+            for part in ref[2:].split("/")
+            if part not in ("$defs", "definitions")
+        ]
+        return "__" + "_".join(parts).capitalize()
+    return "__" + sanitize(ref.split("/")[-1]).capitalize()
+
+
+def _unresolved_type_name(ref: str) -> str:
+    """The bare type name to show a user for an unresolvable ``$ref``."""
+    return ref.rstrip("/").split("/")[-1] or ref
+
+
+def _iter_refs(node: Any) -> Any:
+    """Yield every ``$ref`` string in a schema, at any depth."""
+    if isinstance(node, dict):
+        ref = node.get("$ref")
+        if isinstance(ref, str):
+            yield ref
+        for value in node.values():
+            yield from _iter_refs(value)
+    elif isinstance(node, list):
+        for item in node:
+            yield from _iter_refs(item)
+
+
+def _assert_refs_resolve(schema: dict[str, Any]) -> None:
+    """Fail with a user-facing error if any ``$ref`` target is missing."""
+    for ref in _iter_refs(schema):
+        if ref_resolves(ref, schema):
+            continue
+        raise _invalid_schema(
+            f"Type '{_unresolved_type_name(ref)}' could not be resolved. "
+            f"Check that all $ref targets have matching entries in $defs."
+        )
+
+
+def _valid_identifier(name: str, fallback: str) -> str:
+    """Coerce `name` into something usable as a Python class name."""
+    # the generator treats a leading underscore as private and withholds the class
+    cleaned = name.lstrip("_")
+    if not cleaned or cleaned[0].isdigit():
+        cleaned = f"{fallback}{cleaned}"
+    if keyword.iskeyword(cleaned):
+        cleaned = f"{cleaned}_"
+    return cleaned
+
+
+def _root_class_name(title: str) -> str:
+    """The root's class name, preserving the schema's ``title`` where possible."""
+    return _valid_identifier(re.sub(r"[^0-9a-zA-Z_]", "_", title), "Model")
+
+
+def _nested_class_name(name: str) -> str:
+    """A Pascal-case class name for a ``$defs`` entry or an inline object."""
+    parts = [part for part in re.split(r"[^0-9a-zA-Z]+", name) if part]
+    pascal = "".join(part[:1].upper() + part[1:] for part in parts)
+    return _valid_identifier(pascal, "Model")
+
+
+def _generate(schema: dict[str, Any]) -> tuple[dict[str, type], str]:
+    """Generate the model classes for `schema`, plus the name of its root class."""
+    order: list[str] = []
+
+    def record_name(name: str) -> str:
+        # The root is named first, so the first call is the only one whose name
+        # is user-visible; everything after it is an internal class.
+        cleaned = _root_class_name(name) if not order else _nested_class_name(name)
+        order.append(cleaned)
+        return cleaned
+
+    config = GenerateConfig(
+        input_file_type=InputFileType.JsonSchema,
+        base_class=_BASE_CLASS,
+        strict_refs=True,
+        # The generator renders Python source and execs it. Nothing ever reads
+        # that source, so skip black/isort: ~3x faster, and the library is
+        # moving to external formatters being opt-in anyway.
+        formatters=[Formatter.BUILTIN],
+        allow_population_by_field_name=True,
+        custom_class_name_generator=record_name,
+        type_mappings=_FORMAT_PINS,
+    )
+    module_name = f"{_DYNAMIC_MODULE_PREFIX}_gen_{next(_dynamic_module_counter)}"
+    try:
+        models = generate_dynamic_models(
+            schema,
+            config=config,
+            module_name=module_name,
+            # Each conversion gets its own classes, so one caller mutating a
+            # model cannot affect another agent built from an identical schema.
+            cache_size=0,
+        )
+    except AgentStartupError:
+        raise
+    except Exception as exc:
+        raise _invalid_schema(
+            f"The schema could not be converted to a model: {exc}"
+        ) from exc
+
+    if not models:
+        raise _invalid_schema("The schema produced no model.")
+
+    return models, _root_name(schema, models, order)
+
+
+def _referenced_model_ids(models: dict[str, type]) -> set[int]:
+    """The ids of the models that another model reaches through a field."""
+    referenced: set[int] = set()
+    for model in models.values():
+        for field in cast("type[BaseModel]", model).model_fields.values():
+            for nested in _models_in(field.annotation):
+                # Recursion makes a model reach itself; that does not make it
+                # somebody else's child.
+                if nested is not model:
+                    referenced.add(id(nested))
+    return referenced
+
+
+def _models_with_root_properties(
+    schema: dict[str, Any], models: dict[str, type]
+) -> list[str]:
+    """The models carrying exactly the properties the document itself declares."""
+    properties = schema.get("properties")
+    declared = set(properties) if isinstance(properties, dict) else set()
+    if not declared:
+        return []
+    return [
+        name
+        for name, model in models.items()
+        if set(_fields_by_json_name(cast("type[BaseModel]", model))) == declared
+    ]
+
+
+def _root_name(
+    schema: dict[str, Any], models: dict[str, type], requested: list[str]
+) -> str:
+    """The name of the model generated for the document as a whole.
+
+    Not simply the name the root asked for. The generator renames on collision
+    and resolves the root last, so a ``$defs`` entry can win the requested name
+    while the root is renamed around it -- a schema titled ``Root`` alongside
+    ``$defs/Root`` hands back the definition, with the document's own
+    properties and ``required`` silently gone. The requested name is not
+    dependable either way: names are singularized after this callback returns,
+    and it is called more than once for some subschemas.
+
+    So the root is identified by structure -- it carries the document's own
+    properties, and nothing else refers to it -- with the requested names left
+    to break ties between shapes structure cannot separate (an ``allOf`` root,
+    whose properties come from its branches).
+    """
+    candidates = _models_with_root_properties(schema, models) or list(models)
+    referenced = _referenced_model_ids(models)
+    unreferenced = [name for name in candidates if id(models[name]) not in referenced]
+    candidates = unreferenced or candidates
+    for name in requested:
+        if name in candidates:
+            return name
+    return candidates[0]
+
+
+def _models_in(annotation: Any) -> list[type[BaseModel]]:
+    """Every model class reachable inside a type annotation, containers unwrapped."""
+    found: list[type[BaseModel]] = []
+    stack: list[Any] = [annotation]
+    while stack:
+        current = stack.pop()
+        if isinstance(current, type) and issubclass(current, BaseModel):
+            found.append(current)
+        else:
+            stack.extend(get_args(current))
+    return found
+
+
+def _fields_by_json_name(model: type[BaseModel]) -> dict[str, Any]:
+    """Map each JSON property name to its field, honouring generated aliases."""
+    return {(field.alias or name): field for name, field in model.model_fields.items()}
+
+
+def _models_for_property(
+    models: list[type[BaseModel]], json_name: str
+) -> list[type[BaseModel]]:
+    """The models describing property `json_name` of any of `models`."""
+    found: list[type[BaseModel]] = []
+    for model in models:
+        field = _fields_by_json_name(model).get(json_name)
+        if field is not None:
+            found.extend(_models_in(field.annotation))
+    return found
+
+
+def _child_nodes(
+    node: dict[str, Any], models: list[type[BaseModel]]
+) -> Iterator[tuple[Any, list[type[BaseModel]]]]:
+    """Yield each subschema of `node` with the models it describes."""
+    for combiner in _COMBINERS:
+        for sub in node.get(combiner) or []:
+            yield sub, models
+
+    properties = node.get("properties")
+    if isinstance(properties, dict):
+        for json_name, sub in properties.items():
+            yield sub, _models_for_property(models, json_name)
+
+    for keyword_name in ("items", "additionalProperties"):
+        sub = node.get(keyword_name)
+        if isinstance(sub, dict):
+            yield sub, models
+
+
+def _tag_referenced_models(
+    root: type[BaseModel], schema: dict[str, Any], pseudo_module: ModuleType
+) -> None:
+    """Name every model a ``$ref`` points at, and publish it on the module."""
+    # Walks schema and model tree together rather than matching on class names:
+    # the generator renames on collision, so only the structural walk is reliable.
+    visited: set[tuple[int, int]] = set()
+
+    def name_models(models: list[type[BaseModel]], type_name: str) -> None:
+        for model in models:
+            if not hasattr(model, "__uipath_marker_name__"):
+                cast(Any, model).__uipath_marker_name__ = type_name
+            if not hasattr(pseudo_module, type_name):
+                setattr(pseudo_module, type_name, model)
+
+    def visit(node: Any, models: list[type[BaseModel]]) -> None:
+        if not isinstance(node, dict) or not models:
+            return
+        key = (id(node), id(models[0]))
+        if key in visited:
+            return
+        visited.add(key)
+
+        ref = node.get("$ref")
+        if isinstance(ref, str):
+            name_models(models, _definition_type_name(ref))
+            visit(resolve_pointer(schema, ref), models)
+            return
+
+        for sub, sub_models in _child_nodes(node, models):
+            visit(sub, sub_models)
+
+    visit(schema, [root])
+
+
+def _register_models(models: dict[str, type], pseudo_module: ModuleType) -> None:
+    """Publish every generated class on the pseudo-module and re-home it there."""
+    for name, model in models.items():
+        setattr(pseudo_module, name, model)
+        if isinstance(model, type) and issubclass(model, BaseModel):
+            model.__module__ = pseudo_module.__name__
+            setattr(pseudo_module, model.__name__, model)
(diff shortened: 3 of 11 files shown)


src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
```diff
@@ -0,0 +1,474 @@
...
+        return re.sub(r"[^a-zA-Z0-9_]", "_", name)
+
+    if ref.startswith("#/"):
+        parts = [
+            sanitize(part)
+            for part in ref[2:].split("/")
+            if part not in ("$defs", "definitions")
+        ]
+        return "__" + "_".join(parts).capitalize()
+    return "__" + sanitize(ref.split("/")[-1]).capitalize()
```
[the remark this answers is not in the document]
vldcmp-uipath: Not a defect — the precedence here is the other way round. `.capitalize()` is applied to `"_".join(parts)`, and `"__"` is prepended to the *result*, so the underscores never reach `capitalize()`:

```python
return "__" + "_".join(parts).capitalize()
#      ^^^^ prepended after
```

Verified against the actual function, including the case you raised:

| `$ref` | `_definition_type_name` | legacy `resolve_ref_path` |
|---|---|---|
| `#/definitions/job-attachment` | `__Job_attachment` | `__Job_attachment` |
| `#/$defs/Job_attachment` | `__Job_attachment` | `__Job_attachment` |
| `#/$defs/job_attachment` | `__Job_attachment` | `__Job_attachment` |
| `#/$defs/Foo_Bar` | `__Foo_bar` | `__Foo_bar` |
| `#/$defs/Address/$defs/Country` | `__Address_country` | `__Address_country` |

So `job_attachments.py`'s lookup of `"__Job_attachment"` resolves. `TestLookupContracts::test_definition_is_reachable_by_its_marker_name` asserts exactly that, parametrized over both `job-attachment` and `Job_attachment`, against both backends.

On the second point: yes, `capitalize()` lowercases the tail, so `Foo_Bar` becomes `__Foo_bar`. That is deliberate and load-bearing rather than incidental — it reproduces the legacy converter's `resolve_ref_path` byte for byte (right-hand column above). These names are how callers look types up, so the two backends must agree; "fixing" the casing here would make the marker names diverge the moment the feature flag is flipped, and break job-attachment discovery for any multi-segment name. The docstring says this, and I've left it as is.

---
_Generated by [Claude Code](https://claude.ai/code)_

pyproject.toml
```diff
@@ -19,7 +19,8 @@ dependencies = [
     "python-dotenv>=1.0.1",
     "httpx>=0.27.0",
     "openinference-instrumentation-langchain>=0.1.69, <0.2.0",
-    "jsonschema-pydantic-converter>=0.4.0",
+    "datamodel-code-generator>=0.76.0",
+    "jsonschema-pydantic-converter>=0.4.1",
     "jsonpath-ng>=1.7.0",
```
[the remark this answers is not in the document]
vldcmp-uipath: Good catch — this was a real hole. Fixed in 1456458.

`jsonschema` was indeed only available transitively, and the `try`/`except` around the import returned the model *unguarded* after logging a warning, so an install without it would have silently stopped validating `not`, `prefixItems` and empty `enum` rather than failing.

Taking your second option as well as the first:

- `jsonschema>=4.23.0` is now a declared dependency
- the import moved to module level, so a missing dependency fails loudly at import time instead of degrading at conversion time
- the `try`/`except` degradation path is gone — the guard is installed unconditionally whenever the schema needs it
- `types-jsonschema` added to the dev group (alongside the existing `types-protobuf`), so these calls are type-checked rather than silenced with an ignore

`TestUnenforceableConstraints` covers the enforcement itself, parametrized over both backends and including the constraint nested inside an array and behind a `$ref`.

---
_Generated by [Claude Code](https://claude.ai/code)_

vldcmp-uipath:
### `company-research-agent / staging` on `b3c2a39` — runner DNS, not this change

Adding this because the cause differs from the `company-research-agent / alpha` failure in [my earlier note](https://github.com/UiPath/uipath-langchain-python/pull/1055#issuecomment-5492444985) (that one was a provider `UiPathError` inside the LLM call). This one never reaches a model at all — the sample's DuckDuckGo search tool fails to resolve a hostname:

```
ddgs.exceptions.DDGSException: DNSError: error sending request for url
  (https://wt.wikipedia.org/w/api.php?...) > error resolving DNS >
  no records found for Query {
    name: Name("wt.wikipedia.org.nuqmlukprzpe3dsq1l1xiymenf.gx.internal.cloudapp.net."),
    query_type: A }
```

The runner appended its internal Azure search domain to `wt.wikipedia.org`, so the lookup went to `…gx.internal.cloudapp.net` and found nothing. That is runner network configuration reaching the public internet, several layers below anything this PR touches.

Two further reasons it isn't this change:

- **`b3c2a39` is a commit-message-only amend of `f6eed11`** — `git diff f6eed11 b3c2a39` is empty. This sample passed on `f6eed11` with byte-identical code.
- The failing path is `graph.py` → `langchain_community` DuckDuckGo tool → `ddgs`. No schema conversion is involved, and the feature flag added here is off by default, so the conversion path is the one that was already passing.

No fix is being carried in: there is nothing on `main` to port, and adding network retries to a sample suite would widen a PR that is otherwise about schema conversion.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/uipath_langchain/agent/react/_datamodel_code_generator_base.py
```diff
@@ -0,0 +1,24 @@
...
+  ends up ``"forbid"`` there.
+"""
+
+from pydantic import BaseModel, ConfigDict
+
+
+class UiPathDatamodelCodeGeneratorBaseModel(BaseModel):
+    """Base class for models generated from JSON Schema at runtime."""
+
+    model_config = ConfigDict(serialize_by_alias=True, extra="allow")
```
andreitava-uip, line 30: **Blocking — every non-identifier property name breaks tool invocation.**

`serialize_by_alias=True` makes `model_dump()` alias-keyed. LangChain's `BaseTool._parse_input` then does `getattr(result, k)` for each dumped key — i.e. getattr by *alias*. Legacy keeps the JSON name as the field name so alias == field name and getattr resolves; this backend sanitizes the field name, so it does not.

`BaseUiPathStructuredTool._parse_input` only catches `ValidationError`, and its alias fix-up runs after `super()._parse_input(...)`, so the `AttributeError` escapes first.

Measured through `StructuredToolWithArgumentProperties`, schema `{"properties":{P:{"type":"string"}},"required":[P]}`:

| P | legacy | this backend |
|---|---|---|
| `plain` | OK | OK |
| `first-name` | OK | `AttributeError: 'R' object has no attribute 'first-name'` |
| `a b`, `class`, `1st`, `@id`, `x.y`, `Content-Type` | OK | same `AttributeError` |

`Content-Type` and `@odata.type` are routine in connector schemas. `schema`/`copy` only survive because getattr finds the inherited `BaseModel` method.

This is also the case the backend is meant to improve, so it is the primary intended scenario failing. `TestPropertyNaming` misses it because it asserts `model_dump()` and never invokes a tool.

Fix in code we own: build `validated_input` in `BaseUiPathStructuredTool._parse_input` by mapping alias → `getattr(result, python_name)` instead of delegating to the alias-unaware upstream loop.

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
```diff
@@ -0,0 +1,432 @@
...
+    order: list[str] = []
+
+    def record_name(name: str) -> str:
+        # The root is named first, so the first call is the only one whose name
+        # is user-visible; everything after it is an internal class.
+        cleaned = _root_class_name(name) if not order else _nested_class_name(name)
+        order.append(cleaned)
+        return cleaned
+
+    config = GenerateConfig(
```
andreitava-uip, line 185: **Blocking — `format` is remapped to real Python types, which changes what goes on the wire.**

No `type_mappings` / `type_overrides` is set, so the generator's default `format` mapping applies. Legacy renders every `{"type":"string","format":X}` as `Optional[str]`. Measured differences:

- **`password` → `SecretStr`.** `model_dump()` gives `{'pw': SecretStr('**********')}`, `model_dump_json()` gives `{"pw":"**********"}`. `sanitize_dict_for_serialization` calls plain `model_dump()` with no `SecretStr` branch, and the result is passed as `activity_input=` to `invoke_activity_async`. The connector gets the mask, not the credential. Same via `model_dump(mode="json")` in `process_tool.py:65` and `escalation_tool.py:304`.
- **`email` / `idn-email` → `EmailStr`.** `email-validator` is not declared and is not in `uv.lock`, so `create_model` raises `AgentStartupError: ... email-validator is not installed, run 'pip install pydantic[email]'`. Any agent with an email-formatted property fails to start, with a pip command the user cannot act on.
- **`date-time` → `AwareDatetime`.** `"2024-01-02T03:04:05"` is rejected (`Input should have timezone info`) where legacy accepted it; LLMs commonly emit naive ISO. `model_dump()` also returns a `datetime`, so `json.dumps` on the result raises.
- Nine more retypings (`uuid`→`UUID`, `uri`→`AnyUrl`, `binary`→`bytes`, `date`, `time`, `duration`, `decimal`, `ipv4`/`ipv6`, `path`), each rejecting strings legacy accepted.

Related: nested enums. Legacy's `DynamicEnum` derives from `str` and serializes; the generated `class E(Enum)` does not, and `sanitize_dict_for_serialization`'s `.value` branch only covers top-level values.

Either pin the format mapping back to `str` via `GenerateConfig`, or accept the retyping deliberately — declare `pydantic[email]`, handle `SecretStr` before serialization, and list the changes in the difference table.

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
```diff
@@ -0,0 +1,432 @@
...
+        raise
+    except Exception as exc:
+        raise _invalid_schema(
+            f"The schema could not be converted to a model: {exc}"
+        ) from exc
+
+    if not models:
+        raise _invalid_schema("The schema produced no model.")
+
+    for name in order:
```
andreitava-uip: **Blocking — returns the wrong class when the root `title` collides with a `$defs` name.**

`record_name` appends the *requested* name to `order`; the generator independently renames on collision and the `$defs` class can win the original name, with the root becoming `<Name>Model`. This loop then matches the recorded name and returns the definition.

```python
{"type":"object","title":"Root",
 "properties":{"r":{"$ref":"#/$defs/Root"}},"required":["y"],
 "$defs":{"Root":{"type":"object","properties":{"x":{"type":"string"}}}}}
```

legacy → fields `['r']`. This backend → fields `['x']`, `required` lost, `model_validate({})` accepted, and the real argument is swallowed as an extra because the base sets `extra="allow"`. So the LLM-facing schema, the validation and the payload are all wrong, silently. Same with no `title` and a `$defs` entry named `Model`.

The fallback below only fires when *no* recorded name is in `models`; here the name is present but bound to the wrong class.

Note the comment at line 260 already says name matching is unreliable because the generator renames on collision — that reasoning applies to the root too. Identify the root structurally, or generate under a name that cannot collide and restore the title afterwards.

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
```diff
@@ -0,0 +1,432 @@
...
+        # The generator renders Python source and execs it. Nothing ever reads
+        # that source, so skip black/isort: ~3x faster, and the library is
+        # moving to external formatters being opt-in anyway.
+        formatters=[Formatter.BUILTIN],
+        allow_population_by_field_name=True,
+        custom_class_name_generator=record_name,
+    )
+    module_name = f"{_DYNAMIC_MODULE_PREFIX}_gen_{next(_dynamic_module_counter)}"
+    try:
+        models = generate_dynamic_models(
```
andreitava-uip, line 199: **Blocking — schema content can make the generated-and-`exec`d source import an arbitrary module.**

`datamodel-code-generator` parses `customTypePath` (`parser/jsonschema.py:857`, a first-class aliased field, no opt-in) and `x-python-type`, and emits them as a real `from <module> import <name>` into the source `dynamic.py` execs. The sanitizers here only normalise *names* and never see these keys; nothing upstream filters schema keys.

Reproduced:

```python
{"title":"T","type":"object",
 "properties":{"a":{"type":"string","customTypePath":"marker_mod.Thing"}}}
# -> marker_mod imported, its top-level code ran
```

and the type-confusion variant:

```python
{"properties":{"a":{"type":"string","customTypePath":"builtins.object"}}}
# -> annotation becomes `object`, arbitrary_types_allowed flipped to True,
#    model_validate({"a": 12345}) succeeds on a declared string
```

Legacy is immune — `Optional[str]`, nothing executed.

The value is constrained to a dotted identifier path, so this is "import an arbitrary named module", not token injection. It still runs that module's top-level code in the agent process, and the type confusion is a validation bypass on its own.

Reachability question I could not settle from the repo: `mcp_tool.py:263` sets `input_schema=tool.inputSchema` straight from the remote MCP server and `static_args.py:233` calls `create_model` on it when `argument_properties` is non-empty. If tenants can connect MCP servers we do not control, this is remote input reaching code execution. Otherwise it is agent-author-to-runtime, which is a lower bar.

Either way: strip or reject `customTypePath` and `x-python-type` before calling the generator.

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
```diff
@@ -0,0 +1,432 @@
...
+        if field is not None:
+            found.extend(_models_in(field.annotation))
+    return found
+
+
+def _child_nodes(
+    node: dict[str, Any], models: list[type[BaseModel]]
+) -> Iterator[tuple[Any, list[type[BaseModel]]]]:
+    """Yield each subschema of `node` with the models it describes."""
+    for combiner in _COMBINERS:
```
andreitava-uip, line 310: **`__uipath_marker_name__` cross-contaminates across combiner branches.**

Each `anyOf`/`oneOf`/`allOf` branch is yielded with the parent's unnarrowed `models` list, so `name_models` tags *every* model in that list with that branch's `$ref` name, `hasattr`-guarded so the first branch wins permanently.

With `result: anyOf[Job_attachment, Note]` plus a sibling `note: $ref Note`:

```
legacy: get_json_paths_by_type(m, "__Job_attachment") -> ['$.result']
this:                                                 -> ['$.result', '$.note']
```

That false positive is not benign — `get_job_attachments` then raises `AgentRuntimeError(OUTPUT_VALIDATION_ERROR, category=SYSTEM)` on valid data, because `job_attachments.py` treats any non-empty value at an attachment path as an `Attachment`.

Reverse the branch order and it becomes a false negative instead: with three branches, all classes end up carrying `__Other`, and a plain `att: $ref Job_attachment` property is never found, so attachments are silently skipped.

Multi-`$ref` combiners are the standard shape zod/TypeScript emit for MCP tool schemas, and nothing upstream normalises them.

Narrow the model list per branch, or skip tagging when a branch is a `$ref` and `len(models) > 1` (which is what legacy effectively does for the ambiguous case).

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
```diff
@@ -0,0 +1,432 @@
...
+            model.__module__ = pseudo_module.__name__
+            setattr(pseudo_module, model.__name__, model)
+
+
+def _unenforceable_constraints(
+    schema: dict[str, Any],
+) -> list[tuple[tuple[str, ...], dict[str, Any]]]:
+    """Locate subschemas using a keyword the generator cannot express as a type."""
+    found: list[tuple[tuple[str, ...], dict[str, Any]]] = []
+    seen: set[int] = set()
```
andreitava-uip, line 374: **Shared `$defs` entries lose their constraint guard after the first reference.**

`seen` is keyed on `id(node)` for the whole walk, and a `$ref`'d `$defs` node is the same object each time. The second occurrence short-circuits before `_is_unenforceable`, so no validator is registered for it.

```python
{"properties":{"a":{"$ref":"#/$defs/N"},"b":{"$ref":"#/$defs/N"}},
 "$defs":{"N":{"type":"object","properties":{"n":{"not":{"type":"string"}}}}}}
```

`{"a":{"n":"str"}}` rejected on both; `{"b":{"n":"str"}}` rejected on legacy, **accepted** here. Same for a `$ref` reached from array items, and for one Python dict reused in two places.

Sharing a `$defs` entry across properties is the normal reason `$defs` exists, so this is the most likely of the guard gaps to be hit.

The memo is genuinely needed for `test_recursive_definition`. Keying by traversal path does **not** work — I tried it and it recurses forever, since the path never repeats. A per-branch ancestor set (frozenset of in-progress node ids, not a global set) terminates on both self-recursive and mutually-recursive schemas and fixes this case.

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
```diff
@@ -0,0 +1,432 @@
...
+
+def _is_unenforceable(node: dict[str, Any]) -> bool:
+    """Whether `node` states a rule no Pydantic annotation can carry."""
+    if any(name in node for name in _UNENFORCEABLE_KEYWORDS):
+        return True
+    # An empty enum permits nothing, which no annotation expresses either.
+    return node.get("enum") == []
+
+
+def _child_paths(
```
andreitava-uip, line 405: **`_child_paths` misses keyword positions that `_child_nodes` covers, so guarded keywords silently lapse there.**

This walker enumerates `properties`, dict-form `items` and the combiners. `_child_nodes` (line 250) additionally descends `additionalProperties`. `_unenforceable_constraints` uses only this one.

```
additionalProperties value = {"type":"string","enum":[]}   legacy rejects | here accepted
additionalProperties value with `not`                      legacy rejects | here accepted
additionalProperties value with `prefixItems`              legacy rejects | here accepted
root-level additionalProperties with `not`                 legacy rejects | here accepted
draft-7 tuple form  items: [ {...not...} ]                 legacy rejects | here accepted
```

Adding `additionalProperties` here alone is not enough — I checked both naive variants. Without a path segment the guard validates the map itself against the value schema and falsely rejects `{"meta": {}}`; with a `"{}"` segment but no change to `_values_at_path` the path never resolves and the guard silently no-ops. Both functions have to change together, mirroring the existing `"[]"` handling.

The tuple-`items` gap needs the same treatment in both walkers.

More generally: two hand-rolled walkers over the same document, with different keyword sets and different memo keys, is why this and the two comments above are three separate bugs. One shared context-carrying traversal would close all three and stop the next keyword from leaking.

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
```diff
@@ -0,0 +1,432 @@
...
+        for sub in node.get(combiner) or []:
+            yield sub, path
+
+
+def _build_constraint_guard(
+    constraints: list[tuple[tuple[str, ...], dict[str, Any]]],
+) -> Any:
+    """Return a callable enforcing constraints the generated types cannot carry."""
+    validators = [
+        (path, jsonschema.Draft202012Validator(subschema))
```
andreitava-uip, line 428: **The guard builds validators from detached subschemas and never checks them, so two classes of input escape as raw exceptions.**

1. No `check_schema`, and `enforce` only converts validation *results*. Anything raised while interpreting a malformed subschema propagates raw out of the `mode="before"` validator — all three of these build fine and blow up at `model_validate`:

```
{"a":{"not":"banana"}}                        -> AttributeError: 'str' object has no attribute 'items'
{"a":{"not":{"type":"string","pattern":"["}}}  -> re.error: unterminated character set
{"a":{"not":{"type":"nope"}}}                  -> jsonschema.exceptions.UnknownType
```

None is a `ValidationError`, so `BaseUiPathStructuredTool._parse_input` does not classify it and the job record ends up as `ERROR_AttributeError` / category `UNKNOWN` with no tool or schema context — the opposite of the fail-loudly-at-startup policy stated in `create_output_model`'s docstring.

2. The validator is constructed from the isolated node with no registry rooted at the document, so a `$ref` *nested inside* a guarded subschema cannot resolve and raises `PointerToNowhere` on **every** validation, valid input included. `_unenforceable_constraints` resolves a `$ref` encountered as the node itself but appends nodes containing nested refs verbatim.

Call `check_schema` at build time and raise `AgentStartupError`, build the validator against the whole document (or a `referencing` registry rooted at `schema`), and wrap `iter_errors` so any residual `jsonschema` exception becomes a `ValueError`.

Separate, same function: `enforce` returns early when `data` is not a dict, so a root-level array schema is never guarded at all — `{"type":"array","items":{..."not"...}}` with `[{"n":"str"}]` is rejected by legacy and accepted here.

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
```diff
@@ -0,0 +1,432 @@
...
+from pydantic import BaseModel, model_validator
+
+from uipath_langchain.agent.exceptions import AgentStartupError, AgentStartupErrorCode
+
+from ._datamodel_code_generator_base import UiPathDatamodelCodeGeneratorBaseModel
+from ._schema_refs import ref_resolves, resolve_pointer
+
+# Prefix for the per-conversion pseudo-modules that let qualified-name lookups
+# resolve each schema's generated classes.
+_DYNAMIC_MODULE_PREFIX = "jsonschema_pydantic_converter._dynamic"
```
andreitava-uip, line 49: The prefix and the counter are duplicated from `_legacy_converter.py:32,34` — same prefix string, two independent `itertools.count()` both starting at 0. The Nth module from each backend collides in `sys.modules`, and the later registration replaces the earlier one:

```
dmcg#1   -> ..._dynamic_1
legacy#1 -> ..._dynamic_0
legacy#2 -> ..._dynamic_1     # collision
# sys.modules["..._dynamic_1"].T is now legacy#2's class, not dmcg#1's
```

That is exactly the failure the per-conversion module exists to prevent, per the docstrings in both files.

It needs both backends to run in one process, so production reachability depends on whether flags are ever reconfigured after startup — I could not establish that. It is guaranteed in the test process, where the fixture flips the flag per test.

Move the prefix, counter and `_create_dynamic_module` into one module both backends import.

andreitava-uip (changes requested):
Reviewed the code-generator backend by running both backends over the same schemas and comparing observable behaviour. The 0.4.1 pin is fine — I reproduced the issue on 0.4.0 in a throwaway venv and confirmed it gone on 0.4.1. The pin alone fixes it and needs no code change.

The second backend does not hold the equivalence the description claims. Flipping the flag changes observable behaviour in several ways that are not in `TestBackendDifferences`, and two of them break tool execution outright. Everything below is flag-on only, so the default path is unaffected and this is not a merge risk today — but it blocks enabling the flag anywhere.

The suite does not catch any of it: `tests/agent` passes with `UIPATH_FEATURE_EnableDatamodelCodeGeneratorConverter=true` except `TestBackendSelection::test_defaults_to_the_legacy_backend`, which asserts the default.

Blocking:
- non-identifier property names break every tool call
- `format: password` sends `**********` instead of the value
- a `$defs` name matching the root `title` returns the wrong model
- `customTypePath` in a schema imports and executes an arbitrary module

Also worth noting: `FeatureFlags.configure_flags({})` merges rather than clears — only `reset_flags()` restores the default. The `schema_backend` fixture gets this right; ad-hoc probes are easy to get wrong.

Details inline.

vldcmp-uipath:
Thanks — this is a genuinely good review. I re-derived six of the findings from scratch in a local venv rather than take them on trust, and **all six reproduced exactly as described**. Measured, legacy vs code-generator, flag flipped via `FeatureFlags`:

| finding | legacy | code-generator |
|---|---|---|
| `customTypePath: builtins.object` | `Optional[str]`, `{"a": 12345}` rejected | `object \| None`, **`{"a": 12345}` accepted** for a declared string |
| `format: password` | `model_dump()` → `{'pw': 's3cret'}` | **`{'pw': SecretStr('**********')}`**; `model_dump_json()` → `{"pw":"**********"}` |
| property `first-name` / `Content-Type` | getattr-by-alias OK | **`AttributeError: 'R' object has no attribute 'first-name'`** |
| root `title` colliding with a `$defs` name | fields `['r']`, required `['r']`, `{}` rejected | **fields `['x']`, required `[]`, `{}` accepted** |
| `format: email` | `Optional[str]` | **`AgentStartupError`** at conversion |
| shared `$defs` entry, second reference | both rejected | `a` rejected, **`b` accepted** |

So I'm not going to argue with any of it. Three points I want to concede explicitly rather than leave implied:

**The equivalence claim in the description was mine and it was wrong.** "The whole conversion contract suite runs against both backends" was true as a statement about parametrization and worthless as evidence — your observation that `TestPropertyNaming` asserts `model_dump()` and never invokes a tool is exactly the hole. The suite proved the two backends agree on what I thought to check, which is not the same as agreeing. The "Differences that remain" table is materially incomplete and I'm correcting it.

**Non-identifier property names failing is the worst of these**, because repairing those names is the reason the backend exists. The headline feature is the broken one.

**Your `customTypePath` reachability question I also can't settle from the repo.** `mcp_tool.py:263` does pass `tool.inputSchema` through unfiltered, so if tenants can attach MCP servers we don't control, schema content reaches a `from <module> import <name>` in `exec`d source. Someone with the tenancy model in view should answer that; either way the filter is warranted, and the type-confusion variant is a validation bypass on its own merits.

**On direction.** The `>=0.4.1` bump is uncontested, you verified it independently, and it needs no code change — but `changes_requested` blocks it along with everything else. Rather than hold the actual fix behind a backend that needs real design work (one context-carrying traversal replacing the two walkers, a schema-key filter, a decision on the `format` mapping policy, and the alias-aware `_parse_input` fix in code we own), I'd propose **splitting**: land the pin on its own, and iterate on the code-generator backend in a follow-up against your findings.

Tell me if you'd rather it all stayed here and I'll work through the list in this PR instead.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/uipath_langchain/agent/react/_datamodel_code_generator_base.py
vldcmp-uipath: Confirmed and fixed in `6a28ea6`. I reproduced it through a real `BaseUiPathStructuredTool` first — `AttributeError: 'SendRequest' object has no attribute 'Content-Type'`, and your reading of the cause is exactly right: `result_dict = result_v2.model_dump()` is alias-keyed, then `validated_input[k] = getattr(result, k)` at `langchain_core/tools/base.py:842` looks the value up by alias.

**I fixed it one level lower than you proposed** — `__getattr__` on `UiPathDatamodelCodeGeneratorBaseModel` resolves a declared JSON name to its sanitized field — rather than rebuilding `validated_input` in `_parse_input`. Two reasons, and I'm happy to switch if you disagree:

- Rebuilding `validated_input` means owning upstream's string-input handling, `InjectedToolCallId` injection and default application too, and keeping all of it in sync with LangChain. That's a standing maintenance cost for what is really a lookup mismatch.
- The mismatch isn't specific to `_parse_input` — anything that dumps and reads keys back with `getattr` hits it. And the framing that convinced me: **the legacy backend exposed the JSON name as an attribute**, because the JSON name *was* the field name. This base class exists to make generated models observationally equivalent to legacy ones, so accepting the alias restores the missing half of that contract rather than inventing new behaviour.

Your `_parse_input` fix-up stays, and it's still load-bearing: `schema`/`copy`/`validate` resolve through *normal* lookup to the inherited `BaseModel` method, so `__getattr__` is never consulted and only the explicit fix-up corrects them. You were right that they "only survive because getattr finds the inherited method" — they were returning a bound method as the argument value, and the fix-up is what makes them correct.

On the test gap, you were right too, and it's the part I'm least happy about. `TestPropertyNaming` asserted `model_dump()` and stopped one call short of the failure. There's now `test_tool_invocation_delivers_the_original_name`, which builds a real tool and invokes it for ten names (`Content-Type`, `@odata.type`, `first-name`, `a b`, `x.y`, `1st`, `class`, `schema`, `copy`, `plain`) against both backends. I checked it actually bites: reverting the fix fails 7 of the 10 code-generator variants and leaves every legacy variant green.

Verified no regression to the rest of the base-model contract — `extra="allow"` still preserves extras, a genuinely missing attribute still raises `AttributeError`, and `additionalProperties: false` still forbids.

Leaving this thread open for your call on the approach. The other findings are untouched so far.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
vldcmp-uipath: overall it is a good observation considering the scenario with the MCP, but it should not be marked as blocking. even if this happens, it would include an user-owned module 

also the type-confusion path is a very exaggerated, over-engineering scenario

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
vldcmp-uipath: Taken the first option — the mapping is pinned back, in `c3eb04e`. Derived from the generator's own table rather than listed, so a format added upstream is covered without a change here:

```python
_FORMAT_PINS = [
    f"{base}+{name}={base}"
    for base, formats in json_schema_data_formats.items()
    for name, type_ in formats.items()
    if type_ is not formats.get("default")
]
```

then `type_mappings=_FORMAT_PINS` on the existing `GenerateConfig(...)`. 41 pins across `string`, `integer` and `number`; `boolean`/`object`/`array` have no divergent formats.

**Two corrections to your list, both in the direction of it being worse than described.**

It isn't only `string`. `integer+date-time` → `datetime`, `number+decimal` → `Decimal`, `number+time-delta` → `timedelta`, `number+time` → `time`. Pinning only string formats would have left those behind, which is why the comprehension iterates every base type.

And `ulid` is a second undeclared dependency alongside `email` — it doesn't retype, it crashes:

```
ModuleNotFoundError: No module named 'ulid'
```

On the other hand, one correction the other way: the nine you listed as "each rejecting strings legacy accepted" mostly *accept* the well-formed string. Only `date-time` (via `AwareDatetime`) and the two email formats actually reject. The real damage on the rest is on the way out — `model_dump()` returns `UUID`, `Decimal`, `IPv4Address`, `PosixPath`, `bytes`, `date`, `AnyUrl`, and eight of them then raise `TypeError: Object of type X is not JSON serializable` through `sanitize_dict_for_serialization`. Same conclusion, different mechanism, and worth recording accurately since it's the `json.dumps` path rather than validation that breaks.

New test `TestFormatKeyword::test_format_does_not_change_the_value` asserts the round-trip value and that the dump survives `json.dumps`, over 14 `(type, format, value)` cases against both backends. Asserting the value rather than the annotation is deliberate — it's what catches `password` coming back as a mask. Reverting the pin fails 13 of the 14 code-generator cases (only `integer+int64` survives, an `int` either way) and leaves every legacy case green.

Pinning also means `pydantic[email]` and `ulid` don't need declaring, and there's no `SecretStr` unwrapping to add at the dump sites.

Since this landed via the difference table too: the retyping was never part of what this backend is for, so the table now has nothing to say about `format` — both backends treat it as an annotation.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
vldcmp-uipath: Confirmed and fixed in `7a8d14e`. Reproduced your schema exactly — `fields=['x']`, `required` gone, `model_validate({})` accepted — and both of your variants (`title` colliding, and untitled against a `$defs` entry named `Model`).

**Taking your first suggestion: the root is identified structurally.**

One correction to the mechanism, in your favour — it is worse than "name matching is unreliable". The name is not a handle on the class at all. I instrumented the callback:

```
schema: title "Root" + $defs/Root
asked for : ['Root', 'Root']
built     : {'Root': <the definition>, 'RootModel': <the root>}
```

The generator asks for the root's name **first** but reserves it **last**, so the definition holds `Root` and the root is renamed around it (`RootModel`, not `Model`). And two more ways the requested name diverges from the class that gets built, which rule out repairing the lookup by de-duplicating names ourselves:

- **names are singularized after the callback returns** — an array property `xs` is asked as `xs` and built as `x`
- **the callback is called more than once for some subschemas** — a two-branch `anyOf` property asks for its name twice, so the recorded list is not one entry per class either

So `_root_name` finds the model that carries the document's own declared properties and that no other model reaches through a field. **Neither test is sufficient alone**, which is why they are intersected rather than chained:

| shape | reference graph | property match |
|---|---|---|
| collision + an unreferenced spare `$defs` | 2 candidates | resolves it |
| `allOf` root (properties come from the branches) | resolves it | no candidates |
| collision where the definition declares the same property names | resolves it | 2 candidates |
| root in a `$ref` cycle + an unreferenced spare | wrong candidate | resolves it |

Either filter coming back empty falls through to the other instead of narrowing to nothing, and the requested names are demoted to a tie-break. 32 lines of code across three functions; the old `len(models) == 1` and `"Model" in models` fallbacks are subsumed.

**Worth recording, because it hides itself:** where the colliding definition happens to declare the *same property name* as the root, the returned model looks correct and only the lost `required` gives it away. `TestRootNameCollision` asserts properties **and** `required` for that reason — asserting properties alone passes on the broken code for that case.

**What the user sees**, on the shape that makes this likely — a wrapper object whose `title` matches the entity in `$defs`, which is what zod emits for MCP tools:

```
shown to the model:  { subject, body }   required: [subject]
actually declared:   { ticket }          required: [ticket]
```

so the model calls `create_ticket(subject=...)`, that validates cleanly against the wrong class, and the connector receives a payload with no `ticket` key. Anything sending the correct `{"ticket": {...}}` is rejected with `AgentRuntimeError: Revise the agent prompt or tool configuration so the generated arguments match the tool input schema` — unactionable, since the prompt and the schema are both right. Your "silently" holds where the definition has no required fields, as in your example; where it has some, it is loud but misdiagnosed.

Ten shapes measured before and after: the four collision variants changed, and the other six — unreferenced `$defs`, root array, `allOf` root, `additionalProperties`-only root, self-recursion, plain — are byte-identical. Reverting the fix and keeping the tests fails 5 of 12, code-generator only, every legacy variant green. `mypy`, `ruff` and the full suite clean.

**One residual difference I have left alone**, flagging rather than deciding: the returned class is now the right one but in a collision it is still *named* `RootModel`, so the difference table's "root title" row holds only without a collision. Since the root is asked for first, reserving its name in the callback would push the definition to `Root1` and let the title win — but that trades a name the model sees for a name the model sees, so it did not seem obviously right to do unasked. Say the word if you want it.

Also worth noting for the other guard findings: `$ref: "#"` is rejected by **both** backends at the façade, so the recursion-through-the-document-root shapes are not reachable — which is why a per-branch ancestor set is the right shape for r3905238493 and a global memo is not.

---
_Generated by [Claude Code](https://claude.ai/code)_

andreitava-uip (approved):

src/uipath_langchain/agent/react/_datamodel_code_generator_converter.py
vldcmp-uipath: Correction to my previous reply: it named `7a8d14e` as carrying this fix, and that commit did not. The amend that produced it rewrote only the commit message — the edited files were never staged, so `7a8d14e` was `c3eb04e`'s code with a longer message, and the green CI on it says nothing about this fix. My description of the mechanism and the reasoning stands; the commit reference was wrong.

The fix is now actually in, in **`964341a`**, rebased onto `main` (`89adb80`) to clear the `uv.lock` conflict.

Re-validated on the rebased tree rather than carried over, since the rebase brought a `datamodel-code-generator` bump to **0.76.1** with it — the naming behaviour this fix reasons about is exactly the kind of thing a patch release can move, so it needed re-checking rather than assuming:

- `TestRootNameCollision` — 12 passing against 0.76.1
- `mypy --config-file pyproject.toml .` clean over 380 files, `ruff check` / `ruff format --check` clean
- full suite **3024 passed, 3 skipped, 0 failed**

The `uv.lock` conflict was resolved by taking `main`'s lockfile and re-running `uv lock`, not by editing it. Net effect against `main` is 10 added package entries and none removed: `datamodel-code-generator`, `jsonschema`, `types-jsonschema`, plus the generator's own transitive deps (`black`, `isort`, `genson`, `inflect`, `argcomplete`, `more-itertools`, `pytokens`, `typeguard`), and `jsonschema-pydantic-converter` moving 0.4.0 → 0.4.1. Nothing unrelated re-resolved.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1056, opened 2026-09-01
fix(deps): keep uipath and uipath-runtime ranges within one minor

radu-mocanu:
Releases `uipath-langchain` 0.16.0.post1: 0.16.0's source, with two dependency ranges narrowed to the minors this build was actually released against.

```diff
-version = "0.16.0"
+version = "0.16.0.post1"
-    "uipath>=2.13.16, <2.15.0",
+    "uipath>=2.14.0, <2.15.0",
-    "uipath-runtime>=0.12.5, <0.14.0",
+    "uipath-runtime>=0.13.0, <0.14.0",
```

## Why

0.16.0 was the release that added support for `uipath` 2.14 and `uipath-runtime` 0.13, but it widened both ceilings while leaving the old floors in place. The published ranges therefore span two minors each, and a resolver is free to pair this build with `uipath-runtime` 0.12 or `uipath` 2.13. Neither combination was released or tested against this source.

Narrowing them to a single minor makes the metadata state what the build actually supports. 0.16.2 later landed the same `uipath-runtime>=0.13.0` tightening on the main line; this applies it to 0.16.0 without the source that arrived in between.

## Branch

Cut from `ca64d4f`, the commit that built 0.16.0, per the open-source hotfix procedure. Base branch `release/uipath-langchain-0.16.0` is that same commit. Not for `main`, which is well past 0.16.x.

Source is byte-identical to 0.16.0, which in turn is byte-identical to 0.15.2 under `src/` — `ca64d4f` changed only dependency metadata, and the one commit between 0.15.2 and 0.15.3 was test-only. That is what makes this usable by a consumer that needs `uipath-runtime` 0.13.x without taking any behaviour change.

## Versioning

0.16.1 and later already exist, so per the procedure this is a post-release of the version being fixed rather than a patch increment. PEP 440 orders it `0.16.0 < 0.16.0.post1 < 0.16.1`, so it does not supersede anything for consumers already on a later patch.

## Verification

- `uv lock` pinned at `uipath-runtime==0.13.2`, resolving `uipath` 2.14.1
- `uv sync --locked --all-extras` clean
- full test suite green, exit 0, no failures

## Publishing

CD is `workflow_dispatch`. Dispatch it against the hotfix branch; the merge into the base branch does not trigger it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,14 +1,14 @@
 [project]
 name = "uipath-langchain"
-version = "0.16.0"
+version = "0.16.0.post1"
 description = "Python SDK that enables developers to build and deploy LangGraph agents to the UiPath Cloud Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
 dependencies = [
-    "uipath>=2.13.16, <2.15.0",
+    "uipath>=2.14.0, <2.15.0",
     "uipath-core>=0.5.29, <0.6.0",
     "uipath-platform>=0.2.15, <0.3.0",
-    "uipath-runtime>=0.12.5, <0.14.0",
+    "uipath-runtime>=0.13.0, <0.14.0",
     "uipath-llm-client>=1.17.1, <1.18.0",
     "langgraph>=1.1.8, <2.0.0",
     "langchain-core>=1.2.27, <2.0.0",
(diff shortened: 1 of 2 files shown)


radu-mocanu:
Superseded by #1058, which holds the 0.15 line instead of moving the consumer to 0.16.0. 0.16.0 is byte-identical to 0.15.3, not 0.15.2, so taking it would still have brought the guardrails BYOG change along.

PR #1058, opened 2026-09-01
fix(deps): move the uipath-runtime range to 0.13 on the 0.15 line

radu-mocanu:
Releases `uipath-langchain` 0.15.2.post1: 0.15.2's source with two dependency ranges moved.

**Draft until `uipath` 2.13.23 is published** (UiPath/uipath-python#1878). `uv.lock` is intentionally not regenerated here, because `uv lock` cannot resolve against a version that does not exist yet.

```diff
-version = "0.15.2"
+version = "0.15.2.post1"
-    "uipath>=2.13.16, <2.14.0",
+    "uipath>=2.13.23, <2.14.0",
-    "uipath-runtime>=0.12.5, <0.13.0",
+    "uipath-runtime>=0.13.0, <0.14.0",
```

## Why

Every published 0.15.x caps `uipath-runtime` below 0.13.0. A consumer that has to stay on this line therefore cannot reach `uipath-runtime` 0.13.x, since there is one `uipath-runtime` in an environment and 0.13.x does not satisfy `<0.13.0`. Published metadata is immutable, so the range needs a new release on the line.

The `uipath` floor moves with it because 2.13.23 is the only release on the 2.13 line that admits `uipath-runtime` 0.13.x. Leaving the floor at 2.13.16 would declare support for combinations that cannot resolve.

Both ranges **move** rather than widen, so each stays within a single minor.

## Compatibility

The source is unchanged from 0.15.2, and it was never released against `uipath-runtime` 0.13.x, so that pairing is verified by running it rather than by a prior release: the full test suite passes against 0.13.2, exit 0, no failures.

## Branch

Cut from `ef7600c`, the commit that built 0.15.2, per the open-source hotfix procedure. Base branch `release/uipath-langchain-0.15.2` is that same commit. Not for `main`, which is well past 0.16.x.

## Versioning

0.15.3 and later already exist, so this is a post-release of the version being fixed rather than a patch increment. PEP 440 orders it `0.15.2 < 0.15.2.post1 < 0.15.3`, so it does not supersede anything for consumers already on a later patch.

## Verification

- full test suite green against `uipath-runtime` 0.13.2, exit 0, no failures
- `uv sync --locked --all-extras` clean in that configuration

## Sequencing

1. Publish `uipath` 2.13.23 (UiPath/uipath-python#1878)
2. Run `uv lock` here, push the lockfile, undraft, merge
3. Dispatch CD against this hotfix branch. Merging into the base branch triggers nothing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,14 +1,14 @@
 [project]
 name = "uipath-langchain"
-version = "0.15.2"
+version = "0.15.2.post1"
 description = "Python SDK that enables developers to build and deploy LangGraph agents to the UiPath Cloud Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
 dependencies = [
-    "uipath>=2.13.16, <2.14.0",
+    "uipath>=2.13.23, <2.14.0",
     "uipath-core>=0.5.29, <0.6.0",
     "uipath-platform>=0.2.15, <0.3.0",
-    "uipath-runtime>=0.12.5, <0.13.0",
+    "uipath-runtime>=0.13.0, <0.14.0",
     "uipath-llm-client>=1.17.1, <1.18.0",
     "langgraph>=1.1.8, <2.0.0",
     "langchain-core>=1.2.27, <2.0.0",
--- a/uv.lock
+++ b/uv.lock
@@ -4453,7 +4453,7 @@ wheels = [
 
 [[package]]
 name = "uipath"
-version = "2.13.16"
+version = "2.13.23"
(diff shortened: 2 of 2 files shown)


radu-mocanu:
`uipath` 2.13.23 is now on PyPI, so `uv.lock` is regenerated and committed. Resolved set: uipath 2.13.23, uipath-runtime 0.13.2, uipath-core 0.5.32, uipath-platform 0.2.15.

- `uv sync --locked --all-extras` clean
- full test suite green, exit 0, no failures

Undrafted — this is ready to merge and dispatch.

ionut-mihalache-uipath (approved):

PR #1065, opened 2026-09-03
Hotfix/uipath langchain 0.16.7.post1

yashwagle1:
Cherry pick #1048 bump the current release version from 0.16.7 -> 0.16.7.post1
CI changes so the the checks run against the branch

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - 'release/**'
     paths:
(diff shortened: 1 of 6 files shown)


.github/workflows/cd.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - 'release/**'
```
JosephMar, line 8: should this just be always on?

JosephMar (approved):

PR #1072, opened 2026-09-07
fix(agent): classify LLM provider 400s as USER with an actionable detail

ionut-mihalache-uipath:
## Problem

An LLM gateway `400` fell into the unclassified branch of `_classify`, so it surfaced as `HTTP_ERROR` / `UNKNOWN` with `detail` set to the HTTP reason phrase — the two words **"Bad Request"**.

That is the single largest contributor to the fleet's two-word failure messages (PC-5002), and it buries causes the customer can actually act on. The motivating case: a job that failed 192/192 runs on `gpt-4.1-mini-e2e-custom` because `max_tokens=65535` exceeded the model's 32768 completion-token limit — a value Agent Builder itself had written into the model settings. The customer saw "Bad Request", categorized Unknown.

## Change

`400` now maps to `LLM_PROVIDER_BAD_REQUEST` / `UiPathErrorCategory.USER`, with a title of "LLM provider rejected the request".

For `detail`, precedence is:

1. The gateway's ProblemDetails `detail` — first-party UiPath text, and more specific, so it still wins.
2. Otherwise a canned message pointing at the agent's model settings (output-token limit, temperature, effort) and noting that the provider's own message is on the LLM call span.

The provider body is deliberately **not** read out into the error. It may carry customer PII, and it is already recorded on the LLM call span, which is tenant-scoped. `USER` is not in `_SHOULD_WRAP_CATEGORIES`, so the canned detail has to stand on its own — tests pin that it isn't prefixed with the generic wrapper.

`_classify` now returns a fourth element, `fallback_detail`. `None` means "fall back to the HTTP reason phrase", so only statuses whose cause we cannot name are left with it.

## What is deliberately not in scope

`404` stays `UNKNOWN`. Every LLM-gateway 404 observed in prod over 30 days was a missing or unreachable model deployment — BYO relay not connected, Azure `DeploymentNotFound`, a retired Bedrock model — which is **Deployment**, not User. It needs its own decision on its own evidence rather than being folded into this change. `test_unclassified_4xx_remains_unknown` documents that and covers 404/408/413/422/429.

## Tests

- 400 → USER + canned detail across four body shapes: vendor envelope, Bedrock envelope, raw edge HTML, empty body.
- The provider body is never quoted back — asserted over `detail`, `str(error)` and `repr(error)`.
- A gateway ProblemDetails `detail` takes precedence over the canned text.
- The canned detail is not wrapped in the generic "An unexpected error occurred" prefix.
- End-to-end through the LLM node for both the new client (`UiPathAPIError`) and the legacy raw `openai.BadRequestError` path.
- Unclassified 4xx statuses stay UNKNOWN.

`uv run pytest tests/agent/test_llm.py tests/agent/react/test_llm_node.py` — 54 passed. ruff check, ruff format, the httpx-client AST linter, and mypy all clean.

## Drive-by

Fixed a stale file reference in the `develop-agent-module` skill: `exceptions/licensing.py` → `exceptions/llm.py`.

https://claude.ai/code/session_01UZ5LHC1Xhp1zBRfYnYENwv


tests/agent/react/test_llm_node.py
```diff
@@ -429,6 +429,59 @@ async def test_legacy_raw_provider_error_is_normalized_and_mapped(self):
         assert info.status == 403
         assert info.code.endswith(AgentRuntimeErrorCode.LICENSE_NOT_AVAILABLE.value)
 
+    @staticmethod
+    def _http_400() -> httpx.Response:
+        """The 400 from job 1fab7e97-...: max_tokens written by Agent Builder."""
```
andreiancuta-uipath: Don't reference job keys in code

src/uipath_langchain/agent/exceptions/llm.py
```diff
@@ -98,13 +213,13 @@ def raise_for_provider_http_error(error: UiPathAPIError) -> NoReturn:
...
-    code, category, title = _classify(status_code, body)
-    detail = error.body.get("detail") if isinstance(error.body, dict) else None
+    code, category, title, fallback_detail = _classify(status_code, body)
+    gateway_detail = body.get("detail") if isinstance(body, dict) else None
 
     raise AgentRuntimeError(
         code=code,
         title=title,
-        detail=detail or error.message or str(error),
+        detail=gateway_detail or fallback_detail or error.message or str(error),
```
andreiancuta-uipath: Couldn't we move `gateway_detail or fallback_detail` into `_classify`? That way `_classify` just returns a detail, and the fallback logic doesn't leak out of it.

src/uipath_langchain/agent/exceptions/llm.py
```diff
@@ -21,6 +21,55 @@
 _LICENSE_ERROR_CODE = 10000
 _LICENSE_TITLE = "license not available"
 
+# A canned, provider-free replacement for the useless HTTP reason phrase. The
+# relayed provider message is deliberately NOT read out of the body:
+# it may carry customer PII, and it is already recorded on the LLM call span,
+# which is tenant-scoped. It has to stand on its own -- USER is not in
+# _SHOULD_WRAP_CATEGORIES, so nothing else is prepended to it.
```
andreiancuta-uipath: Overspecific comment.
The "detail is not read out of the body; it may carry customer PII" motivation belongs in the `if status_code == 400:` block in `_classify`. The rest is not useful.

tests/agent/react/test_llm_node.py
ionut-mihalache-uipath: Dropped. The docstring is now "A gateway 400 as seen in prod: max_tokens above the model's limit." — dc48d7fc.

src/uipath_langchain/agent/exceptions/llm.py
ionut-mihalache-uipath: Done. `_classify` now reads `body["detail"]` itself and returns the final detail, so `raise_for_provider_http_error` just uses it. `None` still means "fall back to the HTTP reason phrase", which is the one piece that has to stay in the caller since it needs `error.message`. Precedence is unchanged — gateway detail wins over the canned 400 text and over the 404 signature details. dc48d7fc.

src/uipath_langchain/agent/exceptions/llm.py
ionut-mihalache-uipath: Trimmed to the PII motivation and moved it onto the `if status_code == 400:` branch; `_BAD_REQUEST_DETAIL` now stands with no comment. dc48d7fc.

tests/agent/react/test_llm_node.py
andreiancuta-uipath: There are still other comments referencing the same job.

andreiancuta-uipath (approved):

tests/agent/react/test_llm_node.py
ionut-mihalache-uipath: Missed one — `tests/agent/test_llm.py:248` had the same job key in the `_MAX_TOKENS_BODY` comment. Removed in 85704856; no `1fab7e97` anywhere in the repo now.

src/uipath_langchain/agent/exceptions/llm.py
ionut-mihalache-uipath: Follow-up: applying the precedence inside each branch pushed `_classify` to cognitive complexity 18 (Sonar limit 15), so a4dbaee8 splits it — `_status_verdict` decides code/category/title/detail per status, `_forbidden_verdict` holds the 403 body check, and `_classify` overlays the gateway detail once at the end. Same precedence, complexity now 5 / 4 / 2.

PR #1087, opened 2026-09-15
fix(mcp): keep the MCP connection on its own task

PopescuTudor:
Agents with 2+ MCP sessions open were finishing a run correctly and then failing at teardown with `ERROR_CancelledError`, so the job showed up as Faulted with the right output already produced. ([PRODEV-1583](https://uipath.atlassian.net/browse/PRODEV-1583))

The transport and `ClientSession` sit in anyio cancel scopes. Those have to be exited by the task that entered them, and LIFO on that task. A caller can't guarantee either — langgraph's `ToolNode` opens the connection inside a tool task (`asyncio.gather`) while `dispose()` runs on the teardown task, and an agent with several servers disposes them oldest-first. Both shapes hit it: the first as `Attempted to exit cancel scope in a different task than it was entered in`, the second as `Attempted to exit a cancel scope that isn't the current task's current cancel scope`. Either way it was caught under `except Exception` on the way out and logged at DEBUG, with the half-closed scope resurfacing later as a `CancelledError` — a `BaseException`, so it escaped the handlers above it too.

A dedicated task now opens the connection and holds it until `dispose()` signals it to close. Enter and exit happen on the same task, so the caller's task and dispose order stop mattering. This covers every consumer of `McpClient`, coded agents included, rather than just the low-code factory — which is why it supersedes UiPath/uipath-agents-python#740.

Two things that came out of review:
- A caller cancelled while the handshake is still in flight cancels the connection task outright instead of signalling it — a signal would only be seen once the handshake returned, up to the 600s transport timeout. `ready` is awaited through `asyncio.shield` so the caller's cancellation doesn't cancel the future, which is how `_close_connection` tells a finished handshake from one in flight.
- Setup errors still surface as themselves: unwinding the transport's task group would otherwise wrap them in a `BaseExceptionGroup`, so the original is kept and handed to the caller like before.

Tests, against the real-server harness, all capturing the client logger at DEBUG and asserting nothing was swallowed during teardown:
- a session opened on a tool task disposes from the teardown task (fails on main — the swallowed scope error is in the log)
- three clients dispose cleanly in creation order (fails on main with the production error)
- a caller cancelled mid-handshake unwinds in well under a second (waited the full handshake before)

Two existing tests asserted on `_connection_stack`, which no longer exists; they exercise the same behaviour through the connection task. `agent/tools/mcp/claude.md` updated where it described the old lifecycle.

Bumps to 0.18.5.

`tests/agent/tools/test_mcp` is fully green. The 5 failures elsewhere in the suite are on main too (symlink test needs admin on Windows, a `mimetypes` registry quirk, CLI schema examples).


src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -236,35 +237,101 @@ async def _initialize_client(self) -> None:
...
+        """Ask the connection task to unwind, and wait for it to finish."""
+        task = self._connection_task
+        self._connection_task = None
+        self._session = None
+        if task is None:
+            self._close_requested = None
+            return
+
+        if self._close_requested is not None:
+            self._close_requested.set()
```
radugheo: This only asks politely. If the caller is cancelled while `_run_connection` is still in `_initialize_session()`, setting the event does nothing until the handshake returns, and the `await task` below blocks for the rest of it. Before this PR the handshake ran inline on the caller's task, so cancellation unwound it right away.

I mirrored these three methods in isolation: with a 2.0s handshake, cancellation took 1.91s. The transport default is `httpx2.Timeout(600)`, so the worst case is a tool task that won't die for ten minutes.

When `ready` isn't done yet, `task.cancel()` rather than just signalling. Two things travel with that: `_run_connection` currently catches `CancelledError` without re-raising, and the `except Exception` below wouldn't tolerate a cancelled task.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -236,35 +237,101 @@ async def _initialize_client(self) -> None:
...
+                    )
+                    read_stream, write_stream = streams
+                    self._session = await connection_stack.enter_async_context(
+                        ClientSession(read_stream, write_stream)
+                    )
+                    await self._initialize_session()
+                except BaseException as error:
+                    setup_error = error
+                    raise
+                ready.set_result(None)
```
radugheo: `ready` can already be done here. Awaiting a future and getting cancelled cancels that future, so if the caller was cancelled during setup this raises `InvalidStateError`.

Harmless as it stands, the stack still unwinds on this task, but it surfaces as `MCP connection ended with an error: invalid state` on line 313, which points away from the real cause. Same run as above reproduces it. `if not ready.done():` before the set.

tests/agent/tools/test_mcp/test_mcp_client_task_ownership.py
```diff
@@ -0,0 +1,56 @@
...
+            client = make_client()
+
+            # asyncio.gather wraps the coroutine in its own Task, the way
+            # langgraph's ToolNode dispatches tool calls.
+            (result,) = await asyncio.gather(client.call_tool("add", {"a": 2, "b": 3}))
+            assert result is not None
+
+            await client.dispose()
+
+            assert client.is_client_initialized is False
```
radugheo: This is the test's only assertion, and it passes on main: there `_client_initialized = False` is set in a `finally` whether or not `aclose()` raised. So the cached shape, the half only this PR can fix, has no regression guard.

The description is honest that only `test_clients_dispose_cleanly_in_creation_order` fails on main, but the module docstring says both shapes are covered. Asserting that nothing was swallowed during teardown (caplog), or that the server saw the session close, would make this one bite.

src/uipath_langchain/agent/tools/mcp/claude.md
```diff
@@ -378,7 +378,8 @@ MCP connections for tool invocations with **two distinct initialization phases**
 ├─────────────────────────────────────────────────────────────┤
 │  Connection State (replaced after session loss)             │
 │  ──────────────────────────────────────────────             │
-│  _connection_stack: AsyncExitStack | None                   │
+│  _connection_task: asyncio.Task | None  # owns the scopes   │
```
radugheo, line 381: The box is updated, but three places further down still describe the exit-stack lifecycle:

- line 731, the section "No `with` Statement for AsyncExitStack", labels `async with AsyncExitStack() as stack:` as "Wrong - exits too early". `_run_connection` now does exactly that, correctly. This is the one that matters, since it tells the next person the new pattern is wrong.
- line 530, the Phase 2 diagram, still says "Fresh connection AsyncExitStack".
- line 747, "`_reinitialize_session()` closes the old connection stack".

src/uipath_langchain/agent/tools/mcp/mcp_client.py
PopescuTudor: Fixed in 9a49269f. `_close_connection` now cancels the task while the handshake is still in flight and only signals once it has completed. `ready` is awaited through `asyncio.shield`, so a cancelled caller no longer cancels the future with it — that is what lets `_close_connection` tell the two cases apart. `_run_connection` re-raises `CancelledError` so the task ends cancelled rather than failed, and the wait moved to `asyncio.wait({task})` so the task's own cancellation doesn't read as ours.

`test_cancelling_the_caller_mid_handshake_unwinds_promptly` pins it: 5s handshake, cancellation returns well under a second. Before the change it waited the full 5.00s.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
PopescuTudor: Guarded with `if not ready.done()` before `set_result`. With `ready` shielded in `_open_connection` the caller's cancellation no longer reaches the future, so the path that produced `invalid state` is closed as well; the guard stays as a cheap backstop.

tests/agent/tools/test_mcp/test_mcp_client_task_ownership.py
PopescuTudor: Confirmed. On main that test passes with `Error during MCP connection cleanup: Attempted to exit cancel scope in a different task than it was entered in` sitting in the DEBUG log — I had read it as the cached shape not reproducing; it did, and was swallowed.

All three tests now capture the client logger at DEBUG and assert nothing was logged as an error during teardown; the tool-task one also asserts the gateway saw a `DELETE`. Run against main's `mcp_client.py`, both the tool-task and the creation-order test fail with that exact message.

src/uipath_langchain/agent/tools/mcp/claude.md
PopescuTudor: All three updated. Section 4 is now "Two Stacks, Two Lifecycles": the base stack stays manual, the connection stack is the `async with` inside `_run_connection`, with why that is the correct shape here. The Phase 2 diagram says the transport and session are opened on a dedicated connection task. Section 5 says `_reinitialize_session()` signals the old connection task and waits for it.

radugheo (approved):

PR #1093, opened 2026-09-16
fix(mcp): keep the MCP connection on its own task (0.16 line)

PopescuTudor:
Backport of #1087 onto the `mcp==1.26` client, for the s201/s202 payload ([PRODEV-1583](https://uipath.atlassian.net/browse/PRODEV-1583)). Standalone hotfix off `release/uipath-langchain-0.16.16`; bumps to **0.16.19** (pyproject + uv.lock) so CD publishes on merge.

Note for #1092: it also claims 0.16.19. Whichever merges second needs a re-bump to 0.16.20 — two lines.

Agents with 2+ MCP sessions open finish the run correctly and then fail at teardown with `ERROR_CancelledError`, so the job shows up as Faulted (ring 1 today runs `release/s201` → `uipath-langchain 0.16.7`, which has this client). The HTTP client, transport and `ClientSession` sit in one `AsyncExitStack`; its anyio cancel scopes must be exited by the task that entered them, LIFO. Callers can't guarantee either: langgraph's `ToolNode` opens the connection inside a tool task while `dispose()` runs on the teardown task, and an agent with several servers disposes them oldest-first. The scope error was caught under `except Exception` and logged at DEBUG, and the half-closed scope resurfaced later as a `CancelledError`.

#1087 doesn't cherry-pick here — the 0.18 client is the `mcp` 2.0 / `httpx2` layout from #1053 — so this is the same design re-implemented on the single-stack 1.26 client: `_run_connection` opens the whole stack and holds it until `dispose()` signals it to close, or cancels it if the handshake hasn't finished. Setup errors still surface as themselves (the transport's task group would otherwise wrap them in a `BaseExceptionGroup`). Recovery is unchanged: it re-runs the handshake on the same session, which is a request over the streams and works from any task.

Tests, against a real `FastMCP` 1.x server, all capturing the client logger at DEBUG and asserting nothing was swallowed:
- a session opened on a tool task disposes from the teardown task — fails on the branch as-is (swallowed scope error)
- three clients dispose cleanly in creation order — fails on the branch as-is with the production `CancelledError`
- a caller cancelled mid-handshake unwinds in well under a second

`tests/agent/tools/test_mcp` green (75). The 3 failures elsewhere (`test_uipath_cli_tool.py::test_schema_examples_exist_in_the_cli`) fail identically on the branch as-is. `ruff` and `mypy` clean. `agent/tools/mcp/claude.md` updated where it described the old lifecycle.

Also verified end-to-end on the R1 stack — `uipath-agents release/s201` (factory as-is, creation-order dispose) + this client + `mcp 1.26`: 3 Dynamic servers with one tool call, 2 Cached servers with both/one called from tool tasks — all clean, 0 swallowed. As-is, the Dynamic case dies with the production `CancelledError` and the Cached cases log the swallowed scope errors.


alessandromuresan (approved):

alessandromuresan (commented):
Reviewed this against #1087 and against `release/uipath-langchain-0.16.16` as-is.

The core mechanism ports faithfully — `_run_connection` / `_close_connection` are logically line-for-line: task-owned stack, `ready` resolved post-handshake, `asyncio.shield(ready)` in the caller, `close_requested.set()` for a finished handshake vs `task.cancel()` for one in flight, `asyncio.wait({task})` rather than `await task`, `setup_error` kept so the transport's task group can't wrap it in a `BaseExceptionGroup`, and the `CancelledError` branch ending the task as cancelled. The structural adaptations look like the right calls for this line too: one stack instead of two (0.16 has no HTTP client reused across connections), and no `_open_connection` / `_close_connection_for_recovery` split, since recovery here re-runs the handshake on the same session and that is task-agnostic.

One finding I think is a real bug: #1087 defends the `_client_initialized and _session is None` state in `_ensure_session`, and that defence isn't ported — it can't be verbatim, since there's no `_open_connection` to reopen with. Details inline on the `finally` block. The rest are smaller parity notes (one defensive assignment, some stale `claude.md`, one dropped assertion).

Leaving this as a comment rather than a change request — the call on severity is yours.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -175,52 +180,128 @@ async def _initialize_client(self) -> None:
...
+                if failure is error:
+                    raise
+                raise failure from error
+            if not ready.done():
+                ready.set_exception(failure)
+            else:
+                logger.debug("MCP connection ended with an error: %s", failure)
+        finally:
+            self._stack = None
+            self._session = None
```
alessandromuresan, line 278: **This breaks an invariant that `_ensure_session` still relies on, and I think it can poison the client permanently.**

On the branch as-is, `_session = None` appears in exactly one place — inside `dispose()`, four lines from `_client_initialized = False` — so `_client_initialized ⟹ _session is not None` always held, and `_ensure_session`'s `return self._session  # type: ignore[return-value]` was safe.

This `finally` runs on *every* exit of the connection task and nulls `_session` without touching `_client_initialized`, so the state is now reachable. #1087 guards exactly it:

```python
async with self._lock:
    if not self._client_initialized:
        await self._initialize_client()
    elif self._session is None:
        # A failed replacement leaves the reusable HTTP client intact.
        # Reopen on the next operation instead of poisoning the client.
        await self._open_connection()

    if self._session is None:
        raise RuntimeError("MCP client initialized without a session")
```

`_ensure_session` is untouched here, so there's no equivalent.

Concrete path: `streamable_http_client` is an `@asynccontextmanager` that yields from inside `async with anyio.create_task_group() as tg` with `tg.start_soon(...)` readers/writers (`streamable_http.py:723-748`). A child task failing *after* the handshake — server restart, dropped connection, i.e. the case `SESSION_ERROR_CODES` retry exists for — cancels the group's scope. That scope was entered on the connection task, so `await close_requested.wait()` raises `CancelledError`, the task ends here, `_session` becomes `None`, and `_client_initialized` stays `True`.

The next `call_tool` then gets `None` out of `_ensure_session` and dies on `AttributeError: 'NoneType' object has no attribute 'call_tool'`. That's not an `McpError`, so `_execute_with_retry` won't recover it, and the flag stays `True` — so every later call fails identically until something calls `dispose()`. The branch as-is would have retried the handshake instead.

Smallest fix that fits this line's shape is probably to clear the flag alongside the session, so `_ensure_session` re-runs full init:

```python
finally:
    self._client_initialized = False
    self._stack = None
    self._session = None
    ...
```

(The ordering looks safe for the setup-error path: `ready.set_exception` only schedules the waiter, so `_initialize_client` resumes after this `finally`, and it never sets the flag on that path anyway.)

Worth adding the `if self._session is None: raise RuntimeError(...)` guard from #1087 too, so the `type: ignore` can't hand a `None` to `operation(session)` whatever the cause.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -175,52 +180,128 @@ async def _initialize_client(self) -> None:
...
+        A task still in the handshake would not see the signal until the
+        handshake returned -- up to the transport timeout -- so it is cancelled
+        instead.
+        """
+        task = self._connection_task
+        ready = self._ready
+        close_requested = self._close_requested
+        self._connection_task = None
+        self._ready = None
+        self._close_requested = None
```
alessandromuresan, line 295: #1087 also does `self._session = None` here, *before* the `if task is None: return` below, so the session reference is dropped even when there's no task to wait on.

Not reachable today — `_session` is only ever set inside `_run_connection` and cleared in its `finally`, and `_close_connection` awaits the task — so this is belt-and-braces rather than a bug. But it's the belt-and-braces that keeps the class of problem flagged on the `finally` block non-fatal, and it's two words. I'd port it.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -175,52 +180,128 @@ async def _initialize_client(self) -> None:
...
+            # Shielded so a cancelled caller does not cancel `ready` with it:
+            # _close_connection reads it to tell a finished handshake (signal)
+            # from one still in flight (cancel).
+            await asyncio.shield(ready)
+        except BaseException:
+            await self._close_connection()
+            self._session_info = None
+            raise
 
         self._client_initialized = True
```
alessandromuresan, line 211: Noting the behavioural delta rather than objecting: `_client_initialized = True` now lands *after* the handshake instead of before it, so a handshake failure that comes back as a retryable `McpError` reaches `_reinitialize_session` with the flag `False` and rebuilds the whole connection, where the branch as-is would have re-run `initialize()` on the existing session and transport. Rebuilding seems more correct — the old session's task group is the thing that just errored — and it matches #1087, which also sets the flag after `_open_connection()`. Just flagging it as a real change in retry behaviour for this line.

One knock-on: `claude.md:610` still says *"Set `_client_initialized = True` before calling `_initialize_session()`"*, which this inverts. That one matters more than the other stale doc bits, since it's phrased as a rule for future edits.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -175,52 +180,128 @@ async def _initialize_client(self) -> None:
...
+        fails the same way it did when it opened the stack inline.
+        """
+        close_requested = self._close_requested
+        assert close_requested is not None
+        # Unwinding the transport's task group wraps whatever went wrong in a
+        # BaseExceptionGroup. Callers used to see the original error, so keep it.
+        setup_error: BaseException | None = None
+        try:
+            async with AsyncExitStack() as stack:
+                self._stack = stack
```
alessandromuresan, line 232: `self._stack` is write-only in `src` after this change — assigned here, nulled in the `finally`, and read nowhere (`dispose()` no longer touches it). It survives only because `test_mcp_client.py:525` asserts `session._stack is None` after dispose.

Fine as-is, and the assignment does keep that assertion meaningful. Just calling out that the attribute is now vestigial in case you'd rather drop both it and the assertion — #1087 still has a genuine `_stack` (the reusable HTTP client's) so it isn't a guide here.

src/uipath_langchain/agent/tools/mcp/claude.md
```diff
@@ -500,22 +500,28 @@ async def _reinitialize_session(self) -> None:
             await self._initialize_session()  # Lightweight!
 ```
 
-### 4. No `with` Statement for AsyncExitStack
+### 4. The Connection Lives on Its Own Task
```
alessandromuresan, line 509: This section is updated, but three other spots in the same file still describe the old lifecycle. #1087 updated its equivalent state box; worth the same pass here:

- **lines 185-196** — the state box lists `_http_client` / `_read_stream` / `_write_stream` / `_stack` under *"Client State (created once, reused on session reinit)"* with no `_connection_task`, `_ready` or `_close_requested`. They're no longer reused across reinit either, since the whole stack is per-connection now. #1087's equivalent change was `_connection_stack` → `_connection_task` + `_close_requested`.
- **lines 454-472** — the `_initialize_client` walkthrough still shows `self._http_client = await self._stack.enter_async_context(...)`, i.e. the stack being opened inline on the caller.
- **line 610** — *"Set `_client_initialized = True` before calling `_initialize_session()`"*. Now inverted by the code; see the comment on `_initialize_client`.

Also a smaller one: **line 366** in the init flow diagram still has `4. Calls _initialize_session()` hanging off `_initialize_client`, which is now the connection task's job.

tests/agent/tools/test_mcp/test_mcp_client_task_ownership.py
```diff
@@ -0,0 +1,203 @@
...
+) -> None:
+    """Cached discovery connects on first tool call, i.e. inside a tool task."""
+    caplog.set_level(logging.DEBUG, logger=MCP_CLIENT_LOGGER)
+    with patched_sdk(math_server_url):
+        client = make_client()
+
+        # asyncio.gather wraps the coroutine in its own Task, the way
+        # langgraph's ToolNode dispatches tool calls.
+        (result,) = await asyncio.gather(client.call_tool("add", {"a": 2, "b": 3}))
+        assert "5" in str(result)
```
alessandromuresan, line 157: The three tests mirror #1087's, and inlining the `uvicorn` / `FastMCP` fixture instead of reusing a `real_server` harness makes sense given that harness doesn't exist on this branch.

One assertion didn't come across though: #1087's version of this test also ends with

```python
assert "DELETE" in [r.http_method for r in gateway.records]
```

so nothing here pins that `terminate_on_close` still reaches the server through the new teardown. That path did change shape — the DELETE now goes out while the stack unwinds on the connection task, and it depends on the HTTP client being entered *before* the transport so it's still open when the transport exits. The ordering in `_run_connection` is right, but it's currently unasserted, and a regression would be silent. The inline fixture would need some request recording to cover it.

(Not porting #1087's `test_recovery_continues_when_failed_connection_cleanup_raises` update looks correct — recovery on this line never closes the connection, so there's no equivalent path.)

src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -175,52 +180,128 @@ async def _initialize_client(self) -> None:
...
+        headers: dict[str, str],
+        session_info: SessionInfo,
+    ) -> None:
+        """Hold the HTTP client, transport and session open until asked to close.
+
+        Resolves *ready* once the session is negotiated, so ``_initialize_client``
+        fails the same way it did when it opened the stack inline.
+        """
+        close_requested = self._close_requested
+        assert close_requested is not None
```
radugheo: Can we replace the `assert` with a conditional check? This throws an `AssertionError`, which is misleading.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
```diff
@@ -408,18 +489,7 @@ async def dispose(self) -> None:
...
-                        self._session = None
-                        self._http_client = None
-                        self._read_stream = None
-                        self._write_stream = None
-                        self._session_info = None
-                        self._client_initialized = False
-
+                await self._close_connection()
+                self._session_info = None
+                self._client_initialized = False
```
radugheo: Don't we have to keep the `finally` block? Is there a particular reason it was deleted?

src/uipath_langchain/agent/tools/mcp/mcp_client.py
PopescuTudor: Confirmed and fixed. Reproduced it first: after a successful call, cancelling the connection task (what the transport's task group does when a child fails) and calling again gave `AttributeError: 'NoneType' object has no attribute 'call_tool'` — `test_a_connection_that_dies_on_its_own_is_rebuilt_on_the_next_call` pins it. The task's `finally` now clears `_client_initialized` with the session, so `_ensure_session` re-runs full init on the next call, and it raises `RuntimeError("MCP client initialized without a session")` instead of returning the `type: ignore`d `None`. Also verified end-to-end on the s201 stack: 3 Dynamic servers, and 2 Cached servers with both/one called from tool tasks, all clean.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
PopescuTudor: Ported: `_close_connection` drops `_session` before the `if task is None: return`.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
PopescuTudor: Intended, and now documented as such: the checklist in `claude.md` says the flag is set only after `ready` resolves and is cleared by the task's `finally`, so a handshake that fails with a retryable `McpError` rebuilds the connection on retry rather than re-running `initialize()` on a session whose task group just errored.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
PopescuTudor: Keeping it. On this line it costs one assignment and one null, and it keeps `test_dispose_releases_resources`'s `_stack is None` assertion meaningful without touching an existing test in a hotfix.

src/uipath_langchain/agent/tools/mcp/claude.md
PopescuTudor: All four updated: the state box now lists `_connection_task` / `_ready` / `_close_requested` under a connection-state section owned by the task; the init-flow diagram makes steps 3-4 the task's; the HTTP-client snippet is shown as entered inside `_run_connection`; and the `_client_initialized` rule at the old line 610 is inverted (set only after `ready`, cleared by the task's `finally`).

tests/agent/tools/test_mcp/test_mcp_client_task_ownership.py
PopescuTudor: Added. The inline fixture now wraps the ASGI app to record HTTP methods, and the tool-task test asserts a `DELETE` reached the server after `dispose()` — the HTTP client is entered before the transport, so it is still open when the transport exits and terminates the session.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
PopescuTudor: Removed the `assert`: `close_requested` is now passed into `_run_connection` alongside `ready`, so there is no `Optional` to check.

src/uipath_langchain/agent/tools/mcp/mcp_client.py
PopescuTudor: Restored a `try/finally` in `dispose()`. The connection fields are cleared by the task's own `finally` whatever ends it, so `dispose()` only owns the client-level state (`_session_info`, `_client_initialized`) — the `finally` makes sure a `dispose()` that gets cancelled halfway through `_close_connection()` still leaves the client flagged as disposed.

src/uipath_langchain/agent/tools/mcp/claude.md
```diff
@@ -500,22 +500,28 @@ async def _reinitialize_session(self) -> None:
...
-# ... use stack ...
-await self._stack.__aexit__(None, None, None)
-
-# Wrong - exits too early
 async with AsyncExitStack() as stack:
-    ...  # Stack closes here!
+    ...  # enter HTTP client, transport and session
+    await self._initialize_session()
+    ready.set_result(None)
+    await close_requested.wait()  # dispose() sets this
```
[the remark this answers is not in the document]
PopescuTudor: Updated the "Modifying Client Initialization" checklist: server resolution stays in `_initialize_client()`, resources are entered on the connection task's stack in `_run_connection()`, and `_client_initialized` is set only after `ready` resolves.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-langchain"
-version = "0.16.18"
+version = "0.16.19"
```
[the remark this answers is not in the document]
PopescuTudor: Bumped to 0.16.20 (pyproject + uv.lock) after rebasing onto #1092's merge. This PR is the artifact for `release/s202`, which #744 moved to `uipath-langchain==0.16.19` / `deepagents>=0.7.11` today, so 0.16.19 → 0.16.20 there is a one-patch pin bump. `release/s201` still pins `deepagents==0.5.9` and cannot consume 0.16.19+; it gets the same fix separately as `0.16.7.post2` on `release/uipath-langchain-0.16.7`.

PR #1107, opened 2026-09-22
feat/blocked uip commands

mariadhakalUipath:
Added checks in uipath_cli_tool to block two commands: uip codedagent run / uip function run


mariadhakalUipath:
Closing in favour of a simpler check.

